Issue #378, opened 2011-06-25
Incorrect url generated for versions of different format


jcarlson:
I have an uploader that converts a version to a different format and applies some additional transformations. Initially, everything works, but after loading the model instance from the database, the uploader isn't generating the correct path or url for the version.

Here is my uploader class:

```

class PhotoUploader < CarrierWave::Uploader::Base

  include CarrierWave::RMagick

  storage :file

  # Override the directory where uploaded files will be stored.
  def store_dir
    "uploads/photos/#{model.id}"
  end

  # Large format version, suitable for the standard photo viewer page
  version :large do

    def store_dir
      "uploads/photos/#{model.id}/versions"
    end

    # Resize down to viewer's width and max height
    process :resize_to_fit => [640, 1280]

    # Reduce quality of version for web storage
    process :quality => [50]

    version :reflected do

      process :reflection

      def filename
        super.chomp(File.extname(super)) + '.png'
      end

    end

  end

  private

  def quality(q=50)
    manipulate! {|img| img.write(current_path) {self.quality=q}}
  end

  def reflection
    manipulate!(:format => "png") do |img|
      img = img.wet_floor 0.4, 0.7
      img.write(current_path) {self.quality=50}
    end
  end

end
```

And here's what happens when I reload the instance (using rails console):

```
>> p = Photo.create :image => File.open('/Users/jcarlson/Desktop/DSC_0171.jpg')
=> #<Photo id: 20, release_date: "2011-06-21", image: "dsc_0171.jpg", created_at: "2011-06-25 20:17:58", updated_at: "2011-06-25 20:17:58">
>> p.image.large.reflected.url
=> "/uploads/photos/20/versions/large_reflected_dsc_0171.png"
>> p = Photo.find(20)
=> #<Photo id: 20, release_date: "2011-06-21", image: "dsc_0171.jpg", created_at: "2011-06-25 20:17:58", updated_at: "2011-06-25 20:17:58">
>> p.image.large.reflected.url
=> "/uploads/photos/20/versions/large_reflected_dsc_0171.jpg"
>> 
```

I haven't quite figured out what's going wrong, but clearly the custom filename method on the "reflected" version is not being accounted for. If I check the versions filename, I get something like this:

```
>> p.image.large.reflected.filename
TypeError: can't convert nil into String
    from /Users/jcarlson/Projects/living5to9.com/app/uploaders/photo_uploader.rb:32:in `extname'
[...]

Issue #284, opened 2011-04-18
Incorrect content type for files.


andoriyu:
Every time when I'm trying to upload a file from irb, file has content-type 'binary/octet-stream'. Is there any possible way to force change of content-type to correct one before saving the file?

PR #404, opened 2011-07-17
Rename images when converting formats

jcarlson:
<< new issue based on #403 >>

This should help with situations where a file has multiple versions, and at least one of the versions is of a different format than the master version. This scenario is documented in issue #378 as well as in issue #284 (and possibly others).

Simply pass :rename => true to the existing convert helper method, and the file will be renamed as well as converted. Existing functionality is preserved, i.e., the default is to not rename the image.

When using Fog as the storage provider, you can utilize process :set_content_type to detect the new MIME type and have that relayed to the cloud storage provider.


bengalamx:
+1

PR #509, opened 2011-10-31
Version type.

antonversal:
Allows to set versions type and processing it later. Main purpose is to create versions in background with resque or delay_job.
Example:
       We have uploader with version `:preview` that have type `:delayed`:

``` ruby

    version :thumb do
      process :scale => [200, 200]
    end

    version :preview, :type => :delayed do
      process :scale => [200, 200]
    end

  end
```

When file is cached version :preview will not be created. And we can create it later:

``` ruby
instance.create_versions!(:delayed)
```

also we can create all version for this uploader by passing :all to the method:

``` ruby
instance.create_versions!(:all)
```

and we can create few versions by passing multiple version types to the method:

``` ruby
instance.create_versions!(:delayed, :other_version)
```

Having different versions we can create versions by priorities.

### DIFF
--- a/lib/carrierwave/uploader/versions.rb
+++ b/lib/carrierwave/uploader/versions.rb
@@ -27,9 +27,9 @@ module CarrierWave
         self.version_names = []
 
         attr_accessor :parent_cache_id
-        attr_accessor :create_version_type
+        attr_accessor :create_type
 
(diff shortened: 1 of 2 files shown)


razvanmitre:
is there support now for generating versions on demand?

PR #539, opened 2011-11-23
fix for long url filenames (issue 538)

vrinek:
https://github.com/jnicklas/carrierwave/issues/538


guich-wo:
got same issue. not such an edge case

obliviusm:
me too

PR #691, opened 2012-04-10
Add more sophisticated multi-page PDF support

xtreme-tanzeeb-khalili:
Hi,

I needed a way to extract the cover of a PDF as a separate version. 

Unfortunately, the current multi-page PDF support in the RMagick processor lets you either generate all pages as separate images or one vertically-joined image. 

This patch allows you to accept an additional parameter, zero-indexed page number, from the `manipulate!` block. 

You can then write a custom processor like this:

``` ruby
def cover 
  manipulate! do |frame, index|
    frame if index.zero?
  end
end
```

Single page PDFs and regular images will receive `index = 0`, and existing processors can ignore the index parameter. This patch should also work for other multi-frame files, such as animated GIFs.

### DIFF
--- a/lib/carrierwave/processing/rmagick.rb
+++ b/lib/carrierwave/processing/rmagick.rb
@@ -249,13 +249,14 @@ module CarrierWave
 
       frames = if image.size > 1
         list = ::Magick::ImageList.new
-        image.each do |frame|
-          list << (block_given? ? yield( frame ) : frame)
+        image.each_with_index do |frame, index|
+          processed_frame = block_given? ? yield( frame, index ) : frame
+          list << processed_frame if processed_frame
         end
         block_given? ? list : list.append(true)
       else
         frame = image.first
-        frame = yield( frame ) if block_given?
+        frame = yield( frame, 0 ) if block_given?
         frame
       end
 
--- a/spec/uploader/processing_spec.rb
+++ b/spec/uploader/processing_spec.rb
@@ -83,11 +83,45 @@ describe CarrierWave::Uploader do
       @uploader.process!("test.jpg")
     end
     
-    it "should successfully process a multi-page PDF when using RMagick" do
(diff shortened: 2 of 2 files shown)


davidchristiandy:
I've found that MiniMagick supports layering:

```ruby
def format(format, page = 0)
  @info.clear

  if @tempfile
    new_tempfile = MiniMagick::Utilities.tempfile(".#{format}")
    new_path = new_tempfile.path
  else
    new_path = path.sub(/\.\w+$/, ".#{format}")
  end

  MiniMagick::Tool::Convert.new do |convert|
    convert << (page ? "#{path}[#{page}]" : path)
    yield convert if block_given?
    convert << new_path
  end
  
  ...
end
```

Notice that in the `convert << (page ? "#{path}[#{page}]" : path)` line, it uses an array-like access style. In ImageMagick, using square brackets gives access to specified page. i.e., `convert feb-report.pdf[2] feb-report.jpg` will convert page 2 of the PDF to a JPG.

That said, the obvious thing to do was for me to call `::MiniMagick::Image#format`, and fortunately we can obtain this via blocks in `manipulate!`.

```ruby
version :cover do
  process :convert_cover => :png
end

def convert_cover(format)
  manipulate! do |img| # this is ::MiniMagick::Image instance
    img.format(format.to_s.downcase, 0)
    img
  end
end
```

Now, when I upload a file using this uploader (e.g. using `file.pdf`), it produces `cover_file.pdf`. I thought that this didn't work, but then I noticed the file size difference – 3kb and 55kb. So, tried to manually change the file extension to .png, and it worked. So, the last thing to do is to override the extension. A little bit hardcoding, but it works :smile: 

```ruby
version :cover do
  process :convert_cover => :png
  
  def full_filename (for_file = model.source.file)
    super.chomp(File.extname(super)) + '.png'
  end
end
```

nathanbertram:
Nice solution, thanks @davidchristiandy !

PR #932, opened 2012-12-18
Unscoped model retrieval

div:
I'm using `default_scope` so this find blows up, but i'm not sure that it's a good solution to just go with `unscoped` here. What are your thoughts about it @bensie & @thiagofm ?


adamboxman:
unscoping makes obvious sense. would love to see this issue reopened and merged as this is unintuitive for developers

Issue #1003, opened 2013-02-28
Silent crash on files where the filename contains square brackets "[" or "]"


anandvc:
There's a line in uploader/download.rb which raises an exception within a rescue block due to URI.parse not being able to handle file names that have square brackets "[" or "]" in them. Since the exception occurs within the "rescue" part of the begin-rescue block, no exception message shows up and the operation silently crashes.

PR #1013, opened 2013-03-04
process_uri: handle unencoded brackets

taavo:
Nothing like writing documentation about why a given feature is near impossible to give you ideas on how to implement it relatively elegantly.

We can now handle brackets in filenames. Yes, #1003, I'm talking about you.

How is this possible without being able to parse an invalid URI? Question marks always delimit the query string, so we don't need `URI.parse`—we can just use `split('?')`. Anything before the first question mark should encode brackets. Anything after shouldn't.

(I can also get rid of the known issue I just wrote on this topic. The other characters I was worried about are all fine: colons are permitted, and question marks and hashes always delimit their respective parts of the URL. If there's a problem, it isn't known yet.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -380,9 +380,7 @@ form and you're good to go:
 If you're using ActiveRecord, CarrierWave will indicate invalid URLs and download
 failures automatically with attribute validation errors. If you aren't, or you
 disable CarrierWave's `validate_download` option, you'll need to handle those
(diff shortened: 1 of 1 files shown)


brendon:
A bit late to the party, but I was having problems joining URI's with square brackets in the file name. Using https://github.com/sporkmonger/addressable worked like a charm. Seems like a drop-in replacement.

PR #1148, opened 2013-07-08
Multibyte character sanitize

yalab:
#1141

I would fixed for support ree 1.8.7.


shuhei:
@bensle Hi, do you think it's acceptable now? It's 2015 and carrierwave currently supports 2.0.0 and higher.

bensie:
Definitely. Care to rework and reopen @shuhei?

shuhei:
Sure thing!

PR #1326, opened 2014-02-27
Make the #remove_ accessor set the mounted column as changed

nikz:
This is so that the record is still saved (and the callbacks that do the actual removal are still run) even if no other columns on the model are changed.

This can happen if using `accepts_nested_attributes_for` with a child model that has a carrierwave mounted column, for instance.

### DIFF
--- a/lib/carrierwave/orm/activerecord.rb
+++ b/lib/carrierwave/orm/activerecord.rb
@@ -45,6 +45,11 @@ module CarrierWave
           super
         end
 
+        def remove_#{column}=(value)
+          send(:"#{column}_will_change!")
+          super
+        end
+
         def remove_#{column}!
           super
           _mounter(:#{column}).remove = true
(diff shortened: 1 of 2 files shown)


Awea:
@rmehner it's seems to not be released yet, I get the same problem. Use the version from github in your Gemfile.

phlegx:
@nikz, @rmehner and @Awea I run in the same problem with ```accepts_nested_attributes_for```! The parameter ```remove_avatar``` in nested attributes is submitted, but the avatar still remains.

phlegx:
The master branch has solved this bug: https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/orm/activerecord.rb#L72

@nikz file ```lib/carrierwave/orm/activerecord.rb``` has changed a lot.

nikz:
@phlegx No problems - if everything is fixed in master maybe we can just close this off?

phlegx:
In my opinion, this pull can be closed if a new version, that solves this issue, is released.

PR #1332, opened 2014-03-06
Fixed #public_url to generate proper HTTPS URL.

leods92:
I'm don't know all the ins and outs of CarrierWave and Fog but I know that:
- CarrierWave generates public URLs itself
- Fog also does that in a more thorough way

The problem is CarrierWave public URL generation doesn't consider a HTTPS scenario.
One should not use it with subdomains or custom domains.
Browsers raise security warnings if we do.

That said, I'm suggesting in this PR to use the "directory" URL syntax when HTTPS and custom domains are required.
Anyway, I think this doesn't look DRY and we should leave URL generation to Fog.

What do you guys think?
Maybe I'm missing something here...


edlebert:
@harrisony This is not fixed yet.  You should not generate AWS urls like this: https://my.bucket.name.s3.amazonaws.com/...
You can only do this if your bucket name doesn't have periods.  Or you can do this if using plain http.

harrisony:
@edlebert never said it was :) it's not my pull request, I was simply offering RSpec tests if they were needed

Issue #1337, opened 2014-03-10
Don't require all of Fog


bquorning:
In lib/carrierwave/storage.rb, we currently require all of Fog

``` rb
begin
  require "fog"
rescue LoadError
end
```

And following that, we require lib/carrierwave/storage/fog.rb

Recently (for ~~v1.12.0~~ v1.20.0), Fog was changed so that each storage backend can be required separately: fog/fog#1712.

It would be nice if we could change this to only require `"fog/core"` and let Carrierwave users be responsible for requiring the storage backends they need. In my case, I would just `require 'fog/aws/storage'` somewhere in my application.

blarralde:
This is very useful! Saving almost 10mb of memory with that.

Will a new gem version including that update be released soon? 

PS this is my f4v0r1t3 155u3.

PR #1364, opened 2014-04-11
We don't need to require all of Fog using ~> 1.20

bquorning:
With the release of Fog v1.20.0, it is possible to only require the specific parts of Fog that you need, e.g. `require "fog/aws/storage"` which will load 36 classes, whereas `require "fog"` loads 698 classes in my tests.

See discussion in #1337.

cc @bensie, @elight

### DIFF
--- a/lib/carrierwave/storage.rb
+++ b/lib/carrierwave/storage.rb
@@ -2,7 +2,7 @@ require "carrierwave/storage/abstract"
 require "carrierwave/storage/file"
 
 begin
-  require "fog" unless defined?(Fog)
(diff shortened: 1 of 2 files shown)


webattitude:
hello, is there a stable branche for this ?

bquorning:
I guess [these issues](https://github.com/carrierwaveuploader/carrierwave/milestones/Release%20v1.0.0) have to be fixed before releasing v1.0.0.

bquorning:
But I agree, it would be nice if someone would identify a stable point in the master branch and cut a v0.10.1 or v0.11.0 release.

PR #1446, opened 2014-08-27
Change image extension when converting formats.

nashby:
It's a new version of https://github.com/carrierwaveuploader/carrierwave/pull/404 but I think it should be default behaviour and we don't need that rename option. @bensie what do you think?

### DIFF
--- a/lib/carrierwave/processing/mime_types.rb
+++ b/lib/carrierwave/processing/mime_types.rb
@@ -59,7 +59,7 @@ module CarrierWave
     #
     def set_content_type(override=false)
       if override || file.content_type.blank? || generic_content_type?
-        new_content_type = ::MIME::Types.type_for(file.original_filename).first.to_s
+        new_content_type = ::MIME::Types.type_for(current_path).first.to_s
         if file.respond_to?(:content_type=)
           file.content_type = new_content_type
         else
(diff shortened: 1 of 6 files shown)


mterenzio:
Was this reverted at some point? It wasn't working for me and I notice the relevant code is no longer in Master

dem1tris:
@mterenzio Same question. It must be delegated to `imageprocessing` gem, but now it doesn't work

mshibuya:
@mterenzio @dem1tris Then could you open a new issue? Thanks.

dem1tris:
@mshibuya It seems like #2125, #2126, #2127 are still actual.
As I debugged yesterday, cause is probably in `CarrierWave::Storage::File#cache!`.
Here is `new_file.move_tonew_file.move_to(::File.expand_path(uploader.cache_path, uploader.root), uploader.permissions, uploader.directory_permissions, true)` with the `keep_filename == true`.  

Now `new_path` is formed from uploader's point of view which creates `"foo/bar/version_origname.origext"`. But `new_file` at this moment is already named properly by `imageprocessing`: `"/some/path/origname.newext"`. Although `keep_filename == true` passes `original_filename` as `filename` in hash, file path is rewritten now and ending with old extension again.

Btw, I found the hint in Readme section about RMagick:
```
def filename
      super.chomp(File.extname(super)) + '.newext' if original_filename.present?
end
```
Putting it inside version helped me. But it shouldn't be necessary with MiniMagick.

PR #1456, opened 2014-09-09
Add the cache_only configuration option

jeffkreeftmeijer:
Hey everyone,

I’ve been trying to get test fixtures working on CarrierWave, and came up with [a solution that works by monkey-patching `CarrierWave::Mount::Mounter#store!`](http://jeffkreeftmeijer.com/2014/using-test-fixtures-with-carrierwave/). That way, you prevent CarrierWave from moving the files from the cache directory to the `store_dir`.

It’d be great to be able to set a configuration option in your tests instead of breaking a method in the mounter, so I added a configuration option called “cache_only”, which does the same thing. What do you think?

### DIFF
--- a/lib/carrierwave/uploader/configuration.rb
+++ b/lib/carrierwave/uploader/configuration.rb
@@ -38,6 +38,7 @@ module CarrierWave
         add_config :validate_processing
         add_config :validate_download
         add_config :mount_on
+        add_config :cache_only
 
         # set default values
         reset_config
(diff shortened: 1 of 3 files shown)


pdobb:
@jeffkreeftmeijer How do we use this? Your blog post doesn't how how and I'm not sure how to set the configuration option in Carrierwave in my test_helper. Thanks for any help and for your contribution!

Issue #1480, opened 2014-10-17
Tests are failing


jnicklas:
Due to a bug in fog 1.24.0. I've filed an issue on the fog project here: https://github.com/fog/fog/issues/3204

PR #1481, opened 2014-10-17
Add support for multiple file uploads

jnicklas:
This adds pretty effortless support for multiple file uploads, via a new `mount_uploaders` method (note the plural-s).

All the tests pass when run against fog 1.23.0, but #1480 causes failure against 1.24.0 which is why this is red on Travis.

I've tried to not change any existing behaviour, but this makes one backwards incompatible change. The behaviour of `to_json` has been changed when serializing an uploader which is _not_ mounted on a model. This removes the need to override `serializable_hash` on records, which is super, super hacky and shouldn't have been done in the first place IMO. This drastically simplifies the implementation.

This is a huge change which touches a lot of files and refactors and changes a lot of _very_ old code. As such I'm a bit apprehensive about it, but as a feature this is totally amazing. Drop it in with multiple file upload and it just works.

@bensie et al, please take a look and let me know what you think.

### DIFF
--- a/lib/carrierwave/mount.rb
+++ b/lib/carrierwave/mount.rb
@@ -149,7 +149,7 @@ module CarrierWave
         end
 
         def #{column}_url(*args)
-          _mounter(:#{column}).urls(*args)[0]
+          #{column}.url(*args)
         end
 
         def #{column}_cache
--- a/spec/mount_single_spec.rb
+++ b/spec/mount_single_spec.rb
@@ -200,6 +200,16 @@ describe CarrierWave::Mount do
         @instance.image_url.should be_nil
       end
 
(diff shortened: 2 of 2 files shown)


earksiinni:
For the benefit of future generations who are upgrading from < 1.0.0, I want to point out that @jnicklas' comment about how "The behaviour of `to_json` has been changed when serializing an uploader which is not mounted on a model" is not entirely accurate. This statement is true, but also the changes affect the behavior of `to_json` when serializing an uploader which _is_ mounted on a model. In other words, `to_json`'s behavior (or more accurately, `as_json`'s behavior) is changed no matter what you do.

The changelog states that this PR introduces a breaking change, but the scope of the breakage is much broader than the original comment would lead one to believe.

Issue #1338, opened 2014-03-10
"closed stream" error caused by temp file being deleted


ghost:
I am using CarrierWave to store text files and not images, so this may be an uncommon use case. 

When I upload a file, my `create` action creates an instance of MyModel with a StringIO instance containing the data of my CSV. When I save the instance of MyModel, it looks like the uploader creates a temporary file in /tmp, then writes the file to s3 using fog, then deletes the file in /tmp. 

Here's the issue:
My `create` action doesn't redirect to a `show` or `edit` page, it renders a view itself, so it is reusing the CarrierWave uploader instance. 

A call to mymodel.uploader.read will read from the file in the CarrierWave::Storage::Fog::File object, the `file` method reads from the @file variable. The problem is that this @file variable is memoized and still contains the tempfile when I try to do mymodel.uploader.read in the view rendered by my create action, even though the file has been uploaded to s3.

https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/storage/fog.rb#L383

I have poorly monkeypatched my CarrierWave to fix this immediate problem, but would love other solutions:

``` ruby
class CarrierWave::Uploader::Base
  def read
    # TODO: fix CarrierWave so the file doesn't return a temp file, the problem 
    # is somewhere around
    # https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/storage/fog.rb#L383
    #
    # We have to read from data_file.url because data_file.read is still bound
    # to a temp file in some cases, which has already been deleted, causing a 
    # "closed stream" error. So, if the result of .read isn't a string, we try
    # to treat it like a url!
    content = file.read
    content = open(url).read unless content.is_a? String
    content
  end
end
```

tboyko:
I'm still seeing this problem in 0.11.2. I'm using fog to connect to S3 and have carrierwave mounted to a model attribute in Rails. obj.reload doesn't help. I have to Obj.find() in order to avoid and error when CarrierWave tries to read the temp file that no longer exists.

PR #1517, opened 2014-12-08
Cause Fog::File.read to return its contents after upload

stormsilver:
This solves the case where uploading a file to Fog and then reading its body
immediately afterward results in a deleted, closed tempfile rather
than the actual contents of the upload. #1338


tboyko:
I'm still seeing this problem in 0.11.2. I'm using fog to connect to S3 and have carrierwave mounted to a model attribute in Rails. `obj.reload` doesn't help. I have to `Obj.find()` in order to avoid and error when CarrierWave tries to read the temp file that no longer exists.

PR #1529, opened 2014-12-12
Streaming of SanitizedFile.

TreyE:
Added streaming, where available, for moving/copying of SantizedFile.

This doesn't completely fix the issue with streaming, as that will require a re-work of the existing API to support chunked reads.  However, it does keep large files out of memory where possible (e.g. Streaming a large POST body as binary) - at least in the behaviour of sanitized file.

I'm aware there is a fairly bad hack to support the legacy api - but I can't see a way around it without major incompatibilities.

### DIFF
--- a/lib/carrierwave/sanitized_file.rb
+++ b/lib/carrierwave/sanitized_file.rb
@@ -344,6 +344,15 @@ module CarrierWave
       end
     end
 
+    # HACK: This exists only until all clients implement proper
+    #       IO-like behaviour where the standard read(args)
+    #       method is respected.
+    #       This will allow us to move over slowly to a new streaming
+    #       api without requiring a re-write of exising plugins. 
+    def allows_buffered_read?(io_like)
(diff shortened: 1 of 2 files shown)


bensie:
@TreyE Thanks for this. How would you rework the existing API to properly support this?

bensie:
@TreyE If you do come back to this, please merge in the master branch, you'll have a better time with Travis testing :)

TreyE:
The way I would re-work the existing file is to allow the File api to accept the standard IO read API.  Right now many of them only accept read with 0 arguments.

bensie:
@TreyE We would welcome a patch for that if you're up for it :)

kleyy:
+1

thomasfedb:
@TreyE I'm closing this for now, but feel free to open a new PR if you'd like to attempt your proposed solution of allowing the File API to accept the standard IO #read API.

PR #1535, opened 2014-12-17
Fog storage handles missing files on delete

dblia:
... to mirror the file storage behaviour on delete.

### DIFF
--- a/lib/carrierwave/storage/fog.rb
+++ b/lib/carrierwave/storage/fog.rb
@@ -237,6 +237,7 @@ module CarrierWave
         def delete
           # avoid a get by just using local reference
           directory.files.new(:key => path).destroy
+        rescue ::Fog::Service::NotFound
         end
 
         ##


bensie:
Thanks, but the public API for this method is that success returns true and an error raises.

dblia:
Indeed, it seems that this is the public API for `fog` storage, but shouldn't it
be common with the `file` storage API? Otherwise, migrating from `file` to 
`fog` storage breaks the carrierwave's storage API such as our application. 

Consider the following scenario using `carrierwave 0.10.0` version:

```sh
image = MyImage.last
image.remote_image_url = "http://remote-url-goes-here"
image.save!
```

While the snippet above works properly using `file` storage, migrating to `fog`
raises the following error while trying to save the image to the `fog` provider,
in case the image is uploaded for the first time.

```
Fog::Storage::OpenStack::NotFound: Expected(204) <=> Actual(404 Not Found)
excon.error.response
  :body          => "<html><h1>Not Found</h1><p>The resource could not be found.</p></html>"
  :headers       => {
    "Content-Length"     => "70"
    "Content-Type"       => "text/html; charset=UTF-8"
    "Date"               => "Fri, 06 Mar 2015 12:34:04 GMT"
  }
  :local_address => "A.B.C.D
  :local_port    => 4451
  :reason_phrase => "Not Found"
  :remote_ip     => "X.Y.Z.W
  :status        => 404
from /home/user/vendor/bundle/ruby/1.9.1/gems/excon-0.43.0/lib/excon/middlewares/expects.rb:10:in `response_call'    
from /home/user/bundle/ruby/1.9.1/gems/excon-0.43.0/lib/excon/middlewares/response_parser.rb:8:in `response_call'
from /home/user/bundle/ruby/1.9.1/gems/excon-0.43.0/lib/excon/connection.rb:365:in `response'
from /home/user/bundle/ruby/1.9.1/gems/excon-0.43.0/lib/excon/connection.rb:235:in `request'
from /home/user/bundle/ruby/1.9.1/gems/fog-core-1.27.4/lib/fog/core/connection.rb:63:in `request'
from /home/user/bundle/ruby/1.9.1/gems/fog-1.27.0/lib/fog/openstack/storage.rb:149:in `request'
from /home/user/bundle/ruby/1.9.1/gems/fog-1.27.0/lib/fog/openstack/requests/storage
delete_object.rb:15:in `delete_object'
from /home/user/bundle/ruby/1.9.1/gems/fog-1.27.0/lib/fog/openstack/models/storage/file.rb:46:in `destroy'
from /home/user/bundle/ruby/1.9.1/gems/carrierwave-0.10.0/lib/carrierwave/storage/fog.rb:187:in `delete'
from /home/user/bundle/ruby/1.9.1/gems/carrierwave-0.10.0/lib/carrierwave/uploader/remove.rb:15:in `block in remove!'
[...]
```

So, I believe that different storage providers should expose the same API, in order to
make the transition between them completely transarent to the application. What to 
you think?

eavgerinos:
@bensie After a short conversation with @plribeiro3000 (who happens to be a `fog` maintainer too), I think we should support this seamless transition from file storage to a fog storage.

What do you think? (re-opening for conversation)

plribeiro3000:
I believe that perhaps a rescue from the `fog` exception from inside `carrierwave` seems to be a better idea to me. So `carrierwave` would not depend on that behavior of `fog` since there are many providers and not all of them might be following this pattern.

Perhaps both cases could raise a custom `carrierwave` exception?

thomasfedb:
Hey @eavgerinos, is this PR likely to go ahead? Perhaps otherwise you or @plribeiro3000 could open a new issue addressing fog integration.

plribeiro3000:
@thomasfedb We currently have a related issue in fog/fog#3190. Since this is a `fog` issue i believe we could close this one here and follow that one. What do you guys think?

thomasfedb:
I feel firmly that the API should be consistent between storage backends, whether that be via fog or not. However, this can only go ahead once fog have unified their error handling.

I'm closing this PR. Once fog have unified their approach we can examine this again.

Issue #1127, opened 2013-06-18
Carrierwave with amazon s3 not work with counter_culture gem


shir:
If you create model like this

``` ruby
class UploadTest < ActiveRecord::Base

  belongs_to :category

  attr_accessible :category_id, :something

  mount_uploader :something, PhotoUploader

  counter_culture :category
end
```

with uploader connected to s3 (fog and carrierwave-aws backends have same behavior) after updating file only file name in database is updated, but new file on s3 is not uploaded (old file stays on s3). This issue only on updating file, creating new object works ok.

Not sure if I should create same issue in [counter_culture](https://github.com/magnusvk/counter_culture/issues/new) project.

danielevans:
This issue appears to be caused by the fact that counter_culture is doing a `#dup` and would be fixed by https://github.com/carrierwaveuploader/carrierwave/pull/1544

PR #1544, opened 2015-01-15
Flush mounters on dup of active record model

danielevans:
When an ActiveRecord model is "#dup"ed the mounted carrierwave uploaders retain their reference to the old model.

Gist showing the bug: https://gist.github.com/danielevans/2316c9b07afa7efca92a


danielevans:
This issue appears to be the root cause of #1127

bensie:
Thanks @danielevans!

danielevans:
Thank you!

PR #1568, opened 2015-02-24
Fix ArgumentError when ActiveRecord::Locking::Pessimistic#with_lock call...

yui-knk:
...ed

This method pass one argument to reload, so reload with no args
definition raise ArgumentError.

### DIFF
--- a/lib/carrierwave/orm/activerecord.rb
+++ b/lib/carrierwave/orm/activerecord.rb
@@ -81,7 +81,7 @@ module CarrierWave
         end
 
         # Reset cached mounter on record reload
-        def reload
+        def reload(*)
           @_mounters = nil
           super
         end
--- a/spec/orm/activerecord_spec.rb
+++ b/spec/orm/activerecord_spec.rb
@@ -1490,5 +1490,9 @@ describe CarrierWave::ActiveRecord do
       expect(@event.save).to be_true
       expect(File.exist?(public_path('uploads/old.jpeg'))).to be_true
     end
+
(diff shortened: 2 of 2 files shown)


lib/carrierwave/orm/activerecord.rb
```diff
@@ -81,7 +81,7 @@ def remove_#{column}!
         end
 
         # Reset cached mounter on record reload
-        def reload
+        def reload(*args)
```
amatsuda: Prefer `def reload(*)` since the argument is not used here.

yui-knk: OK, I will change it. Should I rebase commits?

amatsuda: Yep. Please squash the fix commit into the first one.

yui-knk: I see. And `DEPRECATION: `expect { }.not_to raise_error(SpecificErrorClass)` is deprecated. Use `expect { }.not_to raise_error` (with no args) instead`. So I will fix this DEPRECATION :)

amatsuda: Awesome. Is everything ready now?

yui-knk:
Please check [here](https://github.com/carrierwaveuploader/carrierwave/pull/1568/files). Maybe, I fixed to `def reload(*)`.

amatsuda:
Thanks! :sushi:

yui-knk:
:)

Issue #1557, opened 2015-02-02
remove 'require "fog"' from 'lib/carrierwave/storage/fog.rb'


tompesman:
Carrierwave requires the complete fog gem, but it looks like fog is moving all supported cloud providers to separate gems. I've removed the `require "fog"` line from the `lib/carrierwave/storage/fog.rb` file and added `gem 'fog-aws'` to my `Gemfile` which is sufficient for carrierwave's S3 support. I've not made a PR because I cannot test all the cloud providers. I think this is easily tested by someone with the correct setup.

See:
https://github.com/tompesman/carrierwave/commit/08a1232fe2371d25a0ef879bb8a3fb67c6808ef5

bquorning:
The ability to require only selected parts of Fog was implemented in #1364 – check the README https://github.com/carrierwaveuploader/carrierwave#optimized-loading-of-fog. The feature is unfortunately not yet available in a released version of Carrierwave, only on master.

bronislav:
@bquorning It doesn't works with latest version of `fog` gem, because they separated fog gem to the several provider specific gems. And `fog/aws/storage` doesn't exist in the `fog` gem anymore. It is moved to the `fog-aws` gem.

tompesman:
Yes, this is even more optimised :wink: if you include `gem 'fog'` you'll get additional gems of different cloud providers. It looks like the fog gem requires all the extracted gems. There is a new fog-core gem which includes the generic parts which where previously in the fog gem.

gem 'fog'
```
    fog (1.24.0)
      fog-brightbox
      fog-core (~> 1.23)
      fog-json
      fog-radosgw (>= 0.0.2)
      fog-sakuracloud (>= 0.0.4)
      fog-softlayer
      ipaddress (~> 0.5)
      nokogiri (~> 1.5, >= 1.5.11)
    fog-brightbox (0.7.1)
      fog-core (~> 1.22)
      fog-json
      inflecto (~> 0.0.2)
    fog-core (1.27.3)
      builder
      excon (~> 0.38)
      formatador (~> 0.2)
      mime-types
      net-scp (~> 1.1)
      net-ssh (>= 2.1.3)
    fog-json (1.0.0)
      multi_json (~> 1.0)
    fog-radosgw (0.0.3)
      fog-core (>= 1.21.0)
      fog-json
      fog-xml (>= 0.0.1)
    fog-sakuracloud (0.1.2)
      fog-core
      fog-json
[...]

PR #1570, opened 2015-02-24
Do not require all of Fog - works with fog-aws

bquorning:
As mentioned in #1557, Fog is in the process of being split into separate libraries, one per provider.

This means that the [Optimized loading of Fog](https://github.com/carrierwaveuploader/carrierwave#optimized-loading-of-fog) part of the README was no longer working, and that it would be possible to require even fewer parts of Fog if using AWS.


README.md
```diff
@@ -723,10 +723,10 @@ the url to the file on Google.
 
 ## Optimized Loading of Fog
 
-Since Carrierwave doesn't know which parts of Fog you intend to use, it will just load the entire library. If you prefer to load fewer classes into your application, you need to load those parts of Fog yourself *before* loading CarrierWave in your Gemfile.  Ex:
+Since Carrierwave doesn't know which parts of Fog you intend to use, it will just load the entire library (unless you use e.g. fog-aws instead of fog proper). If you prefer to load fewer classes into your application, you need to load those parts of Fog yourself *before* loading CarrierWave in your Gemfile.  Ex:
 
 ```ruby
-gem "fog", "~> 1.27", require: "fog/aws/storage"
+gem "fog", "~> 1.27", require: "fog/google/storage"
```
bensie: Let's keep all the generic examples pointed at AWS.

README.md
```diff
@@ -734,10 +734,10 @@ A couple of notes about versions:
 * This functionality was introduced in Fog v1.20.
 * This functionality is slated for CarrierWave v0.11.
 
-If you're not relying on Gemfile entries alone and are requiring "carrierwave" anywhere, ensure you require "fog/aws/storage" before it.  Ex:
+If you're not relying on Gemfile entries alone and are requiring "carrierwave" anywhere, ensure you require "fog/google/storage" before it.  Ex:
```
bensie: aws here too

README.md
```diff
@@ -734,10 +734,10 @@ A couple of notes about versions:
...
 
 ```ruby
-require "fog/aws/storage"
+require "fog/google/storage"
```
bensie: and here as well

lib/carrierwave/storage.rb
```diff
@@ -2,8 +2,9 @@
 require "carrierwave/storage/file"
 
 begin
-  require "fog" unless defined?(::Fog)
+  require "fog/core"
 rescue LoadError
+  require "fog" unless defined?(::Fog)
```
bensie: Doesn't this require fog as a runtime dependency? Need to still keep it optional.

lib/carrierwave/storage/fog.rb
```diff
@@ -1,6 +1,10 @@
 # encoding: utf-8
 
-require "fog" unless defined?(::Fog)
+begin
+  require "fog/core"
+rescue LoadError
+  require "fog" unless defined?(::Fog)
```
bensie: Same comment here...this will `raise` if fog can't be required.

bensie:
@bquorning Thanks for your work on this! A few comments above, but definitely on board with the idea!

README.md
bquorning: I disagree. If you’re using AWS and want to optimize the loading of Fog, you should use the `fog-aws` gem (as mentioned under Using Amazon S3).

This section (Optimized Loading of Fog) is intended for users of older versions of Fog (v1.20 - v1.27 I guess) or users using a Fog storage library that hasn’t yet been separated from Fog.

lib/carrierwave/storage.rb
bquorning: Yeah, there used to be a `rescue LoadError` after `require "fog"` here. I’m not sure why, but I can add it back.

README.md
bensie: Ah makes sense. :+1:

lib/carrierwave/storage/fog.rb
bquorning: Here, on the other hand, there was no `rescue LoadError` before. And this file is only required in `lib/carrierwave/storage.rb` if `Fog` is defined. So the rescue shouldn’t be necessary.

bquorning:
I added the `resque LoadError`. I’ll squash the commits a bit before it’s ready to merge – but please review :-)

It’s unfortunate that the master build is red. But it doesn’t look like this PR makes any more specs fail.

lib/carrierwave/storage/fog.rb
bensie: Here there was no `rescue LoadError` because if fog isn't available and this file is required, it should raise.

Your `rescue LoadError` is necessary here because we want to load all of fog if it wasn't already loaded by the `require "fog/core"`.

bquorning: After `require "fog/core"`, there *is* a `Fog` defined. (see https://github.com/fog/fog-core/blob/master/lib/fog/core.rb and https://github.com/fog/fog-core/blob/master/lib/fog/core/version.rb)

bensie: At this point I'm not sure if we're saying the same thing :)

Your current code is correct for this file. We need both requires to handle 3 scenarios:
1. You're using an up-to-date version of Fog that has fog/core
2. You're using an older version of Fog that requires you to load the whole thing
3. Fog is not included in the gem bundle and cannot be required, which should `raise` since either 1 or 2 is required to continue.

bensie:
Yeah fixing master is on my list, it's definitely a bummer and really slowing things down :(

bquorning:
Thanks @bensie

tompesman:
@bquorning and @bensie Thanks! :+1:

PR #1579, opened 2015-03-02
Replace double quotes with single in README

stgeneral:
Rails generates Gemfile and configs with a single quotes. So, it's better to follow its probably most populatr convention.


bensie:
Thanks, but this only complicates the git history to figure out where something of substance was last changed.

PR #1596, opened 2015-03-12
Fix #1593 - Magic mime validator

### LINKED ISSUE #1593: [Request For Comments] Magic Mime Validator
This is request for comments for a feature I'm willing to implement for CarrierWave.

There are situations where the _actual_ content-type of a file is important to know and validate, e.g. when processing a file as an image with `minimagick` or when storing a file with metadata in a storage like Openstack's Swift.

Therefore, I was thinking of an optional validator which works similarly to `Extension(Whitelist|Blacklist)`.

A rough skeleton looks like

```Ruby
module CarrierWave::Uploader::MagicMimeWhitelist
  extend ActiveSupport::Concern

  included do
    before :cache, :check_whitelist_pattern!
  end

  # Override in this with a regex
  # example: /image\// for checking images
  def mime_type_whitelist_pattern; end

  private

  def check_whitelist_pattern!
    # do serious stuff here and raise an integrity error if it does not match with the whitelist pattern
  end
end
```

This module is going to use `ruby-filemagic` which is not supported on JRuby, therefore, it must be optional and manually included by the user of the library. Also, a mirror module named `MagicMimeBlacklist` must be provided with the exact opposite logic.

### LINKED ISSUE #1594: avatar? returns false after I remove avatar and then add back gain
Project that demonstrates the problem can be found here: https://github.com/pmatsinopoulos/test_carrierwave

I have a test case in which the `#avatar?` returns `false` where it should be returning `true`.
Here is the test case:

```
require 'test_helper'

class UserTest < ActiveSupport::TestCase
  test "avatar" do
    @user = User.new
    @user.name = "George"
    @user.save!

    assert_equal false, @user.avatar?

    # set avatar
    @user.avatar = File.open("#{Rails.root}/test/files/lion-30-x-30.jpg")
    @user.save!

    assert_equal true, @user.avatar?

    # remove avatar
    @user.remove_avatar!
    @user.save!

    assert_equal false, @user.avatar?

    # add another one
    @user.avatar = File.open("#{Rails.root}/test/files/lion-2-30-x-30.jpg")
    @user.save!

    assert_equal true, @user.avatar?
  end
end
```
Last test fails.

Any clue why?

eavgerinos:
Fixes #1593


bensie:
Looks great to me! Thanks!

Would you mind also adding an entry to the README for its usage?

eavgerinos:
Of course, will do tonight

eavgerinos:
@bensie I cannot find a proper section for it in the README. I think it's better to add a wiki page.

What do you think?

bensie:
@eavgerinos Add a section around https://github.com/carrierwaveuploader/carrierwave#using-minimagick with the other processors.

We tend to say the README is considered official docs and wiki is community-driven, so let's get something for it in the README.

eavgerinos:
Ok :) Thanks for the instant response!

eavgerinos:
@bensie Ready!

bensie:
@eavgerinos Looks great! Thanks!

PR #1598, opened 2015-03-13
doc correction for remove_previously_stored_avatar

somlor:
I noticed that `skip_callback :commit, :after, :remove_previously_stored_avatar` does __not__ prevent previous files from being destroyed on update. 

Using `skip_callback :save, :after, :remove_previously_stored_avatar` instead, does keep previous versions.

See this example app i created to demonstration: https://github.com/somlor/carrierwave_avatar

### DIFF
--- a/README.md
+++ b/README.md
@@ -904,7 +904,7 @@ avatar, even after uploading a new one), you can use ActiveRecord’s
 ```ruby
 class User
   mount_uploader :avatar, AvatarUploader
-  skip_callback :commit, :after, :remove_previously_stored_avatar
+  skip_callback :save, :after, :remove_previously_stored_avatar
 end
 ```
 


bensie:
@somlor I'm a bit confused on this change. From what I can tell, you're triggering `skip_callback` for a callback that doesn't exist.

https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/orm/activerecord.rb#L63

somlor:
@bensie it's possible the issue is more complex than I am realizing, but please clone my example repo and try for yourself: https://github.com/somlor/carrierwave_avatar/blob/master/app/models/user.rb#L15

When `skip_callback :commit, :after, :remove_previously_stored_avatar` is used, updating the user model deletes the previous avatar. When `skip_callback :save, :after, :remove_previously_stored_avatar` is used the previous avatar file is preserved.

thomasfedb:
@somlor What version of Carrierwave are you using? And which version of the documentation are you referring to?

In the master branch, the callback in question is attached to the `:commit` event.
https://github.com/carrierwaveuploader/carrierwave/blob/1e4e7b0bf20fee70d8ac3815d0921144f39e14cf/lib/carrierwave/orm/activerecord.rb#L63

somlor:
Please see my example repo to reproduce this issue. I'm using the latest version: https://github.com/somlor/carrierwave_avatar/blob/master/Gemfile#L6

thomasfedb:
@somlor Yes, you're using the latest **released version**, however you're proposing the change to the **master branch version**, so there might be differences! Can you check if your issues is present in the master branch?

somlor:
@thomasfedb oh nice! I didn't notice this. It appears this may have been the issue all along. I just double checked and `skip_callback :commit, :after, :remove_previously_stored_avatar` seems to be working as expected on master. Mystery solved. Thanks for following up on this. :+1:

thomasfedb:
No worries @somlor. Thanks!

PR #1607, opened 2015-03-25
Update sanitized_file.rb

stevereinke:
Fix mkdir! to handle symlinks

### DIFF
--- a/lib/carrierwave/sanitized_file.rb
+++ b/lib/carrierwave/sanitized_file.rb
@@ -296,7 +296,7 @@ module CarrierWave
     def mkdir!(path, directory_permissions)
       options = {}
       options[:mode] = directory_permissions if directory_permissions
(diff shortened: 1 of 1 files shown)


ardecvz:
Hello,
Could you kindly clarify what you want to solve as it seems there's no need to additionally check for symlink?

`File.exist?` is a simple wrapper around the `stat` system call so it follows symlinks and returns information about the file or directory the link references. `File.symlink?` return different from `File.exist?` result only if it works with a dangling symlink but in this case we're unable to create files in the broken link anyway.

I think it's much better to check if parent is a directory and throw an exception early if it's something different such as a file or symlink.

stevereinke:
Sorry, you are right. I thought File.exist? does not handle symlinks and searching for a solutions seemed to look like that. But you are right, I will cloese this request.

PR #1615, opened 2015-04-02
Another dependency

sethherr:
I needed to run `brew install libmagic` to get `ruby-filemagic` to install.


eavgerinos:
@sethherr Thanks :)

PR #1619, opened 2015-04-08
Small refactors on code

QuotableWater7:
I modified a few files to hopefully make things more clear and/or get rid of duplications.  Let me know what thoughts are on this!

### DIFF
--- a/lib/carrierwave/storage/fog.rb
+++ b/lib/carrierwave/storage/fog.rb
@@ -146,9 +146,7 @@ module CarrierWave
         ).files.all(:prefix => uploader.cache_dir).each do |file|
           time = file.key.scan(/(\d+)-\d+-\d+/).first.map { |t| t.to_i }
           time = Time.at(*time)
-          if time < (Time.now.utc - seconds)
-            file.destroy
-          end
+          file.destroy if time < (Time.now.utc - seconds)
         end
       end
 
@@ -204,8 +202,6 @@ module CarrierWave
             else
               local_file.url(::Fog::Time.now + @uploader.fog_authenticated_url_expiration)
             end
-          else
-            nil
           end
         end
 
@@ -398,9 +394,8 @@ module CarrierWave
         # [NilClass] no file name available
         #
         def filename(options = {})
-          if file_url = url(options)
-            URI.decode(file_url.split('?').first).gsub(/.*\/(.*?$)/, '\1')
-          end
+          return unless file_url = url(options)
+          URI.decode(file_url.split('?').first).gsub(/.*\/(.*?$)/, '\1')
         end
 
         ##


lib/carrierwave/storage/file.rb
```diff
@@ -28,11 +28,8 @@ class File < Abstract
       #
       def store!(file)
         path = ::File.expand_path(uploader.store_path, uploader.root)
-        if uploader.move_to_store
-          file.move_to(path, uploader.permissions, uploader.directory_permissions)
-        else
-          file.copy_to(path, uploader.permissions, uploader.directory_permissions)
-        end
+        strategy = uploader.move_to_store ? 'move_to' : 'copy_to'
+        file.send(strategy, path, uploader.permissions, uploader.directory_permissions)
```
eavgerinos: I don't like calling `send` pointlessly. It skews the Ruby backtrace and we are going to have a bad time during debugging stuff.

lib/carrierwave/uploader/cache.rb
```diff
@@ -118,11 +118,8 @@ def cache!(new_file = sanitized_file)
 
           begin
             # first, create a workfile on which we perform processings
-            if move_to_cache
-              @file = new_file.move_to(File.expand_path(workfile_path, root), permissions, directory_permissions)
-            else
-              @file = new_file.copy_to(File.expand_path(workfile_path, root), permissions, directory_permissions)
-            end
+            strategy = move_to_cache ? 'move_to' : 'copy_to'
+            @file = new_file.send(strategy, File.expand_path(workfile_path, root), permissions, directory_permissions)
```
eavgerinos: same

eavgerinos:
@QuotableWater7 Thanks for contributing, I made you some comments.

lib/carrierwave/storage/file.rb
QuotableWater7: Ok - what do you if I wrap the params in a variable then?

QuotableWater7: Made some updates, let me know what you think!

lib/carrierwave/storage/file.rb
```diff
@@ -28,11 +28,8 @@ class File < Abstract
       #
       def store!(file)
         path = ::File.expand_path(uploader.store_path, uploader.root)
-        if uploader.move_to_store
-          file.move_to(path, uploader.permissions, uploader.directory_permissions)
-        else
-          file.copy_to(path, uploader.permissions, uploader.directory_permissions)
-        end
+        file_params = [path, uploader.permissions, uploader.directory_permissions]
```
eavgerinos: I believe that the existing chunk of code is more readable, after all, multi-line ifs are not so bad.

lib/carrierwave/uploader/cache.rb
```diff
@@ -118,11 +118,8 @@ def cache!(new_file = sanitized_file)
 
           begin
             # first, create a workfile on which we perform processings
-            if move_to_cache
-              @file = new_file.move_to(File.expand_path(workfile_path, root), permissions, directory_permissions)
-            else
-              @file = new_file.copy_to(File.expand_path(workfile_path, root), permissions, directory_permissions)
-            end
+            file_params = [File.expand_path(workfile_path, root), permissions, directory_permissions]
```
eavgerinos: same

eavgerinos:
@QuotableWater7 Please remove the two cryptic changes and squash your commits.

Thanks :)

QuotableWater7:
Would you mind elaborating on why this is more cryptic?

QuotableWater7:
A couple examples of this pattern within CarrierWave:
paperclip.rb L72:  `path ||= File.join(*[store_dir, paperclip_style.to_s, for_file].compact)`
download.rb L29: `super or file.respond_to?(*args)`

This pattern is widely used throughout rails, rspec etc as well.

lib/carrierwave/storage/file.rb
QuotableWater7: @bensie Just curious if you have any thoughts on this bit of code?  I do my best to make things as simple as possible and feel like this code is helping move in that direction.

eavgerinos:
The arguments won't get out of sync between the two calls, there is no code .


Though, I find the usage of splat operator unnecessary, since it adds an extra cpu and memory [overhead](https://gist.github.com/eavgerinos/85ef313f1e7014ead608).
We know the params before invoking the method. Whilst, in `downloader.rb` the number of arguments is

eavgerinos:
I believe, these changes are a matter of taste, and I'm not going to enforce my personal taste here. :)

Though, my opinion is:

The arguments won't get out of sync since there is no other code between those calls and only one of those calls will be executed at a time. 

Also, the splat operator is unnecessary here. In downloader.rb, we use the splat operator because we don't know the exact number of arguments. Splat operator has a good use in a libraries public API.
Here, I believe it only adds an extra cpu and memory [overhead](https://gist.github.com/eavgerinos/85ef313f1e7014ead608).

@QuotableWater7 What do you think?

QuotableWater7:
Hi @eavgerinos,

Sorry - I meant we've got a long list of arguments that are in two places, and making changes means we have to remember to change two spots (and could get out of sync that way).  One could argue it is the tests' responsibility to ensure things happen as intended, so I don't have a terribly strong opinion about that point.

I feel like the CPU/memory by this method call will not be a bottleneck in Carrierwave, and therefore readability is most important.  It seems like we have different styles/preferences here though.

I'll undo the changes - just wanted to have a little discussion :)

Thanks for your review!

eavgerinos:
@QuotableWater7 True. :) It won't be a bottleneck, though I belive we should avoid adding milliseconds of CPU time here and there.

Also, I'm a "tests should ensure behaviour" guy.

Thanks for your contribution.

As I can see you added another commit.

In order to avoid reviewing it again and again, can you ping me when you are done with your changes? :)

eavgerinos:
Oh my bad, I see no new changes. :) Merging, thanks for your time.

PR #1620, opened 2015-04-11
Improve Fog Loading

plribeiro3000:
This PR is an attempt to create an easy support for both providers within `fog` and the ones that were already extracted.
Instead of rely on constant definition to load `fog` support, now it depends on a configuration.
If `fog_provider` is not defined `carrierwave` will try to load `fog` completely. If the user has only `fog-aws` on its Gemfile it will fail.
The use won't need to require any `fog` specifics before `carrierwave` anymore because `carrierwave` will do that now.
The optimized loading of fog is not necessary if the user is using an extracted provider and `google` was extracted so i did some changes on `README` reflecting this.

Let me know if i need to change something else or something is wrong.

### DIFF
--- a/README.md
+++ b/README.md
@@ -607,7 +607,7 @@ You can also pass in additional options, as documented fully in lib/carrierwave/
 
 ```ruby
 CarrierWave.configure do |config|
-  config.fog_provider = 'fog-aws'
+  config.fog_provider = 'fog-aws'                        # required
   config.fog_credentials = {
     provider:              'AWS',                        # required
     aws_access_key_id:     'xxx',                        # required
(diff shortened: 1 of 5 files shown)


bensie:
Awesome, thanks!

plribeiro3000:
Wow. Thank you for the fast support. :heart:

plribeiro3000:
@bensie When will be a new release including this changes?

eavgerinos:
@plribeiro3000 Nice! :) Thanks

plribeiro3000:
@bensie @eavgerinos Did i forgot something in here?
In a second thought i think the tests should break somehow. Or perhaps i should add tests for this new config. Do you guys think it would be good to have tests for this?

thejbsmith:
I was having the same issue, and this fixed it for the most part.

One note: `config.fog_provider = 'fog-aws'` should be `config.fog_provider = 'fog/aws'`. With the directory structure of the `fog-aws` gem, there is no `fog-aws.rb` file, the file is located at `fog/aws.rb`.

Once I got that switched, this seems to work great for me. Probably worth updating the README.

eavgerinos:
@thejbsmith Fixed :)

PR #1625, opened 2015-04-14
Fix #1622

### LINKED ISSUE #1622: Can't use carrierwave without fog after #1620
I'm using carrierwave with https://github.com/sorentwo/carrierwave-aws instead of fog. 
After #1620 CW fails to load when fog is not installed/bundled.

eavgerinos:
This removes the mandatory loading of fog.

@bensie @plribeiro3000 Do you believe we need to add any tests to ensure this won't break again?


plribeiro3000:
:+1: for the changes.
I think that perhaps 1 integration test should be simple enough and easy to make sure there will be no regressions.
Can you do that? Or perhaps i should add it on another PR?

eavgerinos:
@plribeiro3000 I don't have time to craft it today.

You can just push a commit with the test to this branch if you like :)

Thanks

plribeiro3000:
I will see if i can do it today then. =)

plribeiro3000:
@eavgerinos I've been playing with this and i think there is no simple way to test this.
We will need to create 2 sample apps inside `spec` because as we have both `fog` and `carrierwave-aws` listed in the `Gemfile` the bug can't be reproduced because `fog` will be in the load path.

As the fix is really simple perhaps we can leave it without tests if you don't feel strong about it.

As for your proposed fix and i think that perhaps we can make this without force any breaking in the usage for now. 

I was thinking about change [this line](https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/uploader/configuration.rb#L127) and [this one](https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/uploader/configuration.rb#L132):

```ruby
eager_load_fog(value) if '#{name}' == 'fog_credentials'
```

to this:

```ruby
eager_load_fog(value) if '#{name}' == 'fog_credentials' and value.any?
```

This would make `CarrierWave::AWS` work again without any changes on the configuration.

We could set `fog` as default for `fog_provider` in [this block](https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/uploader/configuration.rb#L156-L186). This would create breaking changes only for those using an extracted provider (`fog-aws` and `fog-google`).

What do you think?

eavgerinos:
@plribeiro3000 Thanks for investigating this.

Till weekend it's highly unlikely that I'll have time to fix the specs.

Though, people depend on our master branch and we should fix it as soon as possible.
I agree with your changes, in order to avoid many breaking changes, though, a minor comment, I think `value.present?` fits better, what do you think?

plribeiro3000:
Won't `.present` only check for nil values?
If it does it will still break because the default is an empty hash.
If not i'm ok with it. =)

eavgerinos:
Empty hashes are not `present?`

plribeiro3000:
Yeah, just tested. It will work ok. I was just not sure. =)
It is good to go.

plribeiro3000:
Do you want me to add those changes here?

eavgerinos:
Doing it right now, I'll ping you to review and try it :)

eavgerinos:
@plribeiro3000 Added the one fix.

As for setting `fog` as default `fog_provider` will break my [fix](https://github.com/carrierwaveuploader/carrierwave/pull/1625/files#diff-a2fae088fb0c57593e3c4c6bd69bc382R112)

plribeiro3000:
@eavgerinos Why would it break? (Just wondering)

eavgerinos:
@plribeiro3000 Because it would try to require the `fog` gem, resulting to the initial problem reported in #1622

plribeiro3000:
But if you are checking in the calls for `eager_load_fog` for empty params it would never be called. Or i am missing something?

plribeiro3000:
I think the test are broken because of the api that changed now. So all tests must be updated.
Should we head this way and change the tests?

README.md
```diff
@@ -593,6 +593,16 @@ Processing can be enabled for a single version by setting the processing flag on
 @uploader.thumb.enable_processing = true
 ```
 
+## Fog
+
+If you want to use fog you must add in your CarrierWave initializer the
+following lines
+
+```ruby
+config.fog_provider = 'fog'
```
plribeiro3000: I believe this line is not entirely necessary. `'fog'` is the default so the user might need to specify the config only if it using an extracted provider, like `fog-aws` or `fog-google`.

Perhaps we could add a note about this?

eavgerinos: You are right, I'll fix it today!

eavgerinos: fixed

plribeiro3000: :+1:

eavgerinos:
@plribeiro3000 So, I'm merging?

bensie:
:+1:

plribeiro3000:
I think so. :+1:

PR #1628, opened 2015-04-21
Configuration will break only for extracted providers

plribeiro3000:
cc/ @eavgerinos 

This changes are related to my proposal. This way the actual configuration will break only for the extracted providers of fog (`fog-google` and `fog-aws`). This will make this fix transparent for users of `carrierwave-aws` and users that use `rackspace` from `fog`.

Thoughts?


eavgerinos:
Nice, I'll cherry-pick it to my branch to avoid a pointless merge commit.

So, now AFAICT, fog provider defaults to 'fog' without requiring 'fog' though if `for_credentials` is missing. Am I right?

plribeiro3000:
Yes. Thats the behavior. `fog` will be required only if `fog_credentials` is not empty.
So users of `carrierwave-aws` can use it without issues again. =)

eavgerinos:
Awesome! :) Updating the README in my branch and merging it after tests pass :+1: Good job!

plribeiro3000:
I'm happy to help. Please let me know if we need to improve anything else in the context of `fog`.

As we are moving stuff, this informations can help us drive in the right direction. :smile:

PR #1630, opened 2015-04-22
Move to Rspec 3 syntax and update rspec version

eavgerinos:
### Changelog
 
- Update Rspec gem to version 3.2.0
- Update Cucumber gem to version 2.0.0
- Change all specs syntax to be compatible with the new rspec syntax.

### Rspec reference

- `obj.stub(:method)` is converted to `allow(obj).to receive(:method)`
- `expect(condition).to be_false` is converted to `expect(condition).to be_fausey`
- `expect(condition).to be_true` is converted to `expect(condition).to be_truthy`

---

The rationale behind this change is to have more clear specs and make it easier to maintain and extend our test suite which leads to easier adding/modifying of features.


plribeiro3000:
Looking good so far! Great work. :+1:

eavgerinos:
@bensie @plribeiro3000 Please review it after the build passes :)

eavgerinos:
Fixed cucumber features also. I think it's ready.

bensie:
Looks awesome!

PR #1637, opened 2015-04-29
simplify logic on bool method

QuotableWater7:


eavgerinos:
@QuotableWater7 Thanks

PR #1640, opened 2015-05-02
Clarify what's necessary for mount_uploaders

trevrosen:
Some quick README fixes to try to combat issue thrash:

* Add a line clarifying that 1.0.0 is still forthcoming
* Add a line clarifying how to bundle the gem so that the multiple uploader support referenced in the README is available.


thiagofm:
:+1: this makes sense, today it's a bit misleading.

PR #1618, opened 2015-04-08
fix use rmagick 2.14

satoruk:
https://github.com/gemhome/rmagick/commit/77827346b046c6e42ce27909970036d7cc1f16de

> [DEPRECATION] requiring "RMagick" is deprecated. Use "rmagick" instead

PR #1643, opened 2015-05-06
Get rid of rmagick deprecation warning

BastianHofmann:
Rmagick introduced a  deprecation warning in 2.14.0 for loading it with `require "RMagick"` (https://github.com/gemhome/rmagick/blob/master/ChangeLog#L7)


BastianHofmann:
Looks like #1618 was faster than me. I will close this, but please consider merge the other commit.

PR #1656, opened 2015-06-03
Rename (column -> mount_on)_will_change! in ActiveRecord

nviennot:
All the other methods do that

### DIFF
--- a/lib/carrierwave/orm/activerecord.rb
+++ b/lib/carrierwave/orm/activerecord.rb
@@ -73,7 +73,8 @@ module CarrierWave
         end
 
         def remove_#{column}=(value)
-          send(:"#{column}_will_change!")
+          column = _mounter(:#{column}).serialization_column
+          send(:"\#{column}_will_change!")
           super
         end
 


bensie:
@nviennot Can you provide some additional reasoning for this?

nviennot:
Consistency with the other `XXX_will_change!` calls:

https://github.com/nviennot/carrierwave/blob/master/lib/carrierwave/orm/activerecord.rb#L19-L20
https://github.com/nviennot/carrierwave/blob/master/lib/carrierwave/orm/activerecord.rb#L34-L35
https://github.com/nviennot/carrierwave/blob/master/lib/carrierwave/orm/activerecord.rb#L67-L69
https://github.com/nviennot/carrierwave/blob/master/lib/carrierwave/orm/activerecord.rb#L76-L77

that call in `remove_column` is the only one that doesn't lookup the serialization column. That's suspicious. Either `XXX_will_change!` takes a serialization column name, or a attribute name. But not both.

thomasfedb:
This appears to be a sensible change, and code looks okay. Will however require tests. @nviennot are you able to add tests?

nviennot:
I'm not sure what should be tested.

bensie:
Ah I see now, makes sense. Thanks @nviennot!

thomasfedb:
@bensie Do these changes request tests?

bensie:
Not really, it just makes it consistent with the others like it.

PR #1659, opened 2015-06-04
mount_uploaders: Assigning column to [] fails to remove files

nviennot:
Here's a failing test case.
When assigning `@instance.images = []`, the files are not removed for some reason.

I tried to fix it by removing the condition at https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/mount.rb#L327 -- but some other tests start to fail.

Thanks,
Nico :)

### DIFF
--- a/spec/mount_multiple_spec.rb
+++ b/spec/mount_multiple_spec.rb
@@ -557,6 +557,16 @@ describe CarrierWave::Mount do
         expect(@instance).to receive(:write_uploader).with(:images, nil)
         @instance.write_images_identifier
       end
+
+      it "should remove from the column when files are empty" do
+        expect(@instance).to receive(:write_uploader).with(:images, ["test.jpg"])
+        @instance.images = [stub_file('test.jpg')]
+        @instance.write_images_identifier
+
+        expect(@instance).to receive(:write_uploader).with(:images, [])
+        @instance.images = []
+        @instance.write_images_identifier
+      end
     end
 
     describe '#images_identifiers' do


eavgerinos:
Did you try `@instance.remove_images!; @instance.save` ?

nviennot:
Yeah that should work, though this is not my concern.

This is what happens with a regular model:

```ruby
instance.update!(images: [file1, file2])
instance.reload
instance.images # returns [file1, file2], not surprising

instance.update!(images: [file1])
instance.reload
instance.images # returns [file1], not surprising

instance.update!(images: [file2])
instance.reload
instance.images # returns [file2], not surprising

instance.update!(images: [])
instance.reload
instance.images # returns [file2], very surprising
```

Here's a test case for the ActiveRecord ORM if that helps:

```ruby
it "should update the files accordingly on assignments" do
  @event.update!(:images => [stub_file('old.jpeg'), stub_file('new.jpeg')])
  @event.reload
  @event.images.size.should == 2
  @event.images[0].path.should =~ /old.jpeg/
  @event.images[1].path.should =~ /new.jpeg/

  @event.update!(:images => [stub_file('old.jpeg')])
  @event.reload
  @event.images.size.should == 1
  @event.images[0].path.should =~ /old.jpeg/

  @event.update!(:images => [stub_file('new.jpeg')])
  @event.reload
  @event.images.size.should == 1
  @event.images[0].path.should =~ /new.jpeg/

  @event.update!(:images => [])
  @event.reload
  @event.images.size.should == 0 # fails here
end
```

thomasfedb:
@nviennot it appears that is is not so much a bug, but rather a API design choice.
@eavgerinos, @bensie, what are you thoughts about this?

bensie:
Agreed @thomasfedb, `remove_images!` is the documented API for file removal

nviennot:
I leave unsatisfied with the arguments I'm given for this API decision choice.

thomasfedb:
@nviennot Feel free to submit a few issue with your reasoning as to why the API should be changed. However this is not a bug, and this issue will not be reopened.

PR #1668, opened 2015-06-07
[Fix #1594] Reset remove attribute after invoking remove_#{image}

### LINKED ISSUE #1594: avatar? returns false after I remove avatar and then add back gain

eavgerinos:
This fixes the behaviour of write_#{column}_identifier in order to
accept re-uploading a file after the previous one was removed from
the same Uploader's instance.

### DIFF
--- a/lib/carrierwave/orm/activerecord.rb
+++ b/lib/carrierwave/orm/activerecord.rb
@@ -78,9 +78,10 @@ module CarrierWave
         end
 
         def remove_#{column}!
-          super
           self.remove_#{column} = true
           write_#{column}_identifier
+          self.remove_#{column} = false
+          super
         end
 
         # Reset cached mounter on record reload
--- a/spec/orm/activerecord_spec.rb
+++ b/spec/orm/activerecord_spec.rb
@@ -457,17 +457,22 @@ describe CarrierWave::ActiveRecord do
       before do
         @event.image = stub_file('test.jpeg')
         @event.save!
-        @event.remove_image!
       end
 
       it "should clear the serialization column" do
+        @event.remove_image!
+
         expect(@event.attributes['image']).to be_blank
       end
 
-      it "should return to false after being saved" do
-        @event.save!
-        expect(@event.remove_image).to eq(false)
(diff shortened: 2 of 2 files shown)


eavgerinos:
I'm not sure this should go in this method. Maybe it's a responsibility of `write_#{column}_identifier`

What do you think? @bensie @plribeiro3000

plribeiro3000:
@eavgerinos I'm not sure either but this change its kind of bad. The result of the method will be false when it should reflect if the result of the operation was successful or not.

Maybe we could leave the code as it is if we just pull the `super` to the end of the method. Thoughts?

eavgerinos:
@plribeiro3000 Fixed. :)

plribeiro3000:
:+1:

plribeiro3000:
@eavgerinos If is not to ask too much maybe we could have a test to ensure this won´t break in the future? or at least some documentation?

eavgerinos:
@plribeiro3000 Added :)

plribeiro3000:
Awesome @eavgerinos . You are the best!  :+1:

eavgerinos:
@plribeiro3000 :)

PR #1672, opened 2015-06-09
Shouldn't affect parent class' version (closes #1495)

kuraga:

PR #1674, opened 2015-06-10
Add test: .version with nested versions should set the version name for the #versions

kuraga:
There is this test in features, but there isn't it in unit tests.

By the way, it failed when I was manipulating with #1672, but unit tests were green.


kuraga:
Thanks.

PR #1679, opened 2015-06-16
Fixed StringIO support

Benjamin-Dobell:
Fixed two bugs in this method:

1. Previous implementation was checking ```original_filename``` was non-nil but then using ```full_original_filename``` - these may not be the same.

2. ```full_original_filename``` will return an empty string ("") if the Uploader is uploading a file that is *not* a version and the "file" being uploaded doesn't have a path (i.e. is a StringIO). This means you end up writing to a *file* path like:

```
tmp/cache_id
```

This is fine if your Uploader has no versions attached to it. However, if your Uploader also has versions attached to it they end up trying to copy_to:

    tmp/cache_id/version_cache_name

This fails because ```tmp/cache_id``` is a file, not a directory.

Consequently, this pull request will ensure the main (non-version) upload is written to:

    tmp/cache_id/cache_id

when ```full_original_filename``` is ```blank?```. This avoids the issue whereby we're trying to treat a file like a directory.

### DIFF
--- a/lib/carrierwave/sanitized_file.rb
+++ b/lib/carrierwave/sanitized_file.rb
@@ -161,7 +161,7 @@ module CarrierWave
       else
         @file.rewind if @file.respond_to?(:rewind)
         @content = @file.read
-        @file.close if @file.respond_to?(:close) && @file.respond_to?(:closed?) && !@file.closed?
+        @file.close if @file.respond_to?(:close) && @file.respond_to?(:closed?) && !@file.closed? && !@file.is_a?(StringIO)
         @content
       end
     end
--- a/lib/carrierwave/uploader/cache.rb
+++ b/lib/carrierwave/uploader/cache.rb
@@ -128,7 +128,6 @@ module CarrierWave
           with_callbacks(:cache, new_file) do
             self.cache_id = CarrierWave.generate_cache_id unless cache_id
 
-            @filename = new_file.filename
             self.original_filename = new_file.filename
 
             if move_to_cache
@@ -136,6 +135,8 @@ module CarrierWave
             else
               @file = new_file.copy_to(cache_path, permissions, directory_permissions)
             end
+
+            @filename = @file.filename
           end
         end
       end
--- a/lib/carrierwave/uploader/cache.rb
+++ b/lib/carrierwave/uploader/cache.rb
@@ -94,7 +94,13 @@ module CarrierWave
       # [String] a cache name, in the format YYYYMMDD-HHMM-PID-RND/filename.txt
       #
       def cache_name
-        File.join(cache_id, full_original_filename) if cache_id and original_filename
+        if cache_id
+          if full_original_filename.present?
+            File.join(cache_id, full_original_filename)
+          else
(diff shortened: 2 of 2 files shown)


Benjamin-Dobell:
More commits on this branch, but they're all related i.e. They're all necessary in order to ensure StringIO can be utilised. They all of course stem from the fact that StringIO do not come with a filename.

Also, we shouldn't be closing StringIO, the contents are lost indefinitely when we do so. This is a problem if you try assign the same StringIO to multiple attributes etc.

Benjamin-Dobell:
This commits aren't overly useful on their own, so I'm happy to squash them. I'm intentionally basing these on the 0.10 release, so ideally a 0.10.1 bug-fix release could be made.

joallard:
*crickets*

thomasfedb:
@Benjamin-Dobell Can you please explain what this is aiming to acheive? How does `StringIO` work into Carrierwave usecases?

Benjamin-Dobell:
Been a while, so I couldn't even remember. However, I had a quick look through the source to remind myself...

When a image is provided using syntax like:

```
    remote_<field>_url = "http://someurl.com/someimage.jpg"
```

eventually the following is called:

```
            @file = Kernel.open(@uri.to_s)
```

in `download.rb`. Kernel.open returns a `StringIO` for remote URIs. The issues described in my initial comment then arise from there.

thomasfedb:
Thanks @Benjamin-Dobell, are you able to include a test cases that demonstrates the current issue?

Benjamin-Dobell:
Pfft, tests... haha, yeah, I'll try get around to it in the next day or so.

joallard:
I ended up here because reading SanitizedFile, the doc says it's meant to wrap around namely StringIO

```
  ##
  # SanitizedFile is a base class which provides a common API around all
  # the different quirky Ruby File libraries. It has support for Tempfile,
  # File, StringIO, Merb-style upload Hashes, as well as paths given as
  # Strings and Pathnames.
```

And so I expected `StringIO` to work, in my case I was storing an email attachment obtained as a webhook from Mandrill, where you have a hash of `{filename:, type:, data:}`. As a workaround, I'm now creating Tempfiles.

So probably trying to store a file's content from a StringIO would make for an easy test.

thomasfedb:
@Benjamin-Dobell hey dude, any news on those tests? :smile:

thomasfedb:
@Benjamin-Dobell also, you'll need to `git pull upstream; git merge master`.

thomasfedb:
@Benjamin-Dobell I'd like to get this merged if you're able to provide tests.

thomasfedb:
I'm closing this for now @Benjamin-Dobell.

Benjamin-Dobell:
Sorry I'm presently on my honeymoon.

The behaviour definitely should be corrected though. However, I don't mind in the slightest if someone wants to reimplement it themselves in another commit and merge that.

thomasfedb:
Thanks @Benjamin-Dobell, I'll certainly make sure there's an issue opened. If you do get the chance to update this at some point I'd love to merge it. Cheers.

PR #1683, opened 2015-06-19
Change Google's url to the public_url

m7moud:
`commondatastorage` is a deprecated endpoint, the current [Request URIs](https://cloud.google.com/storage/docs/reference-uris) should be used instead, and `fog-google` is already up to date.

I guess this could be extended to the other providers too.

### DIFF
--- a/lib/carrierwave/storage/fog.rb
+++ b/lib/carrierwave/storage/fog.rb
@@ -353,7 +353,8 @@ module CarrierWave
                 end
               end
             when 'Google'
-              "https://commondatastorage.googleapis.com/#{@uploader.fog_directory}/#{encoded_path}"
+              file = directory.files.get(encoded_path)
+              file.nil?? "" : file.public_url
             else
               # avoid a get by just using local reference
               directory.files.new(:key => path).public_url


lib/carrierwave/storage/fog.rb
```diff
@@ -353,7 +353,8 @@ def public_url
                 end
               end
             when 'Google'
-              "https://commondatastorage.googleapis.com/#{@uploader.fog_directory}/#{encoded_path}"
+              file = directory.files.get(encoded_path)
+              file.nil?? "" : file.public_url
```
webattitude: what is #nil?? ?

webattitude: :$ ok... space is not needed then

webattitude:
this is quite long now, I mean asking fog for the public_url.... I've got plenty of timeouts with this now, occuring withing excon socket, so until I sort all this out I reverted to previous version of this class.

thomasfedb:
@webattitude If you're having issues with timeouts when requesting `public_url` please open a new issue.

PR #1701, opened 2015-07-20
Fixed README for multiple file uploads

AndyObtiva:
Multiple file uploads wouldn't work for me until I had the Controller permit the multi-file-upload-attribute, but not simply, yet pointing to an empty array, so I added the fix to the README.


AndyObtiva:
As per Kuraga's request, made sure README file permissions stay the same and added [ci skip] to the commit description to avoid CI given it is a text file change.

bensie:
Thanks!

PR #1718, opened 2015-08-12
Provide whitelisted/blacklisted types for better error messages in MagicMime

StephenCharles:
Similar to the original `extension_whitelist` and `extension_blacklist` CarrierWave functions, I wanted to provide the whitelisted & blacklisted types to error messages for MagicMime.

The provided whitelist & blacklist functions for MagicMime return a regexp not string extensions so passing it to the error message did not look great. I decided to create two new functions `whitelist_mime_type_extension` and `blacklist_mime_type_extension`. They both return an array of strings that represent the filetypes you are whitelisting or blacklisting in the pattern functions. These are then passed as locals to the corresponding errors.

Then in my `config/locale/en.yml` I could use it like so
```
en:
  errors: 
    messages:
      mime_type_pattern_white_list_error: "cannot be filetype %{content_type}, allowed filetypes: %{allowed_types}"
```

### DIFF
--- a/lib/carrierwave/uploader/magic_mime_blacklist.rb
+++ b/lib/carrierwave/uploader/magic_mime_blacklist.rb
@@ -51,19 +51,31 @@ module CarrierWave
       #
       def blacklist_mime_type_pattern; end
 
+      # === Returns
+      #
+      # [Array] a black list array of extensions to match blacklist regexp, 
+      #         is passed as a local to error message  
+      #
+      # === Examples
+      #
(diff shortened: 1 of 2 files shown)


eavgerinos:
@StephenCharles Thanks for your contribution. :)

IMHO, I believe an implementation with an array, like the one in the extensions, is kinda limiting, since you cannot accept a group of mime-types.

Also, I believe that custom error messages is a thing that should be solved on the application side.

PR #1722, opened 2015-08-18
Passing options to Rackspace authenticated_url

duhast:
Hi!

This update allows to pass additional options to Rackspaces `authenticated_url `.
Options are described [here](http://docs.rackspace.com/files/api/v1/cf-devguide/content/TempURL_File_Name_Overrides-d1e213.html).

Corresponding `fog` PR is fog/fog#3665

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ spec/test.log
 spec/tmp
 *.swp
 .rvmrc
+.idea
 .bundle
 Gemfile.lock
 gemfiles/*.lock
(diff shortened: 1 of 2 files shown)


bensie:
@duhast Does this require bumping to the latest Fog for that method signature to be accepted?

duhast:
@bensie Ideally yes, but in case the current version is used, nothing should break but these special Rackspace options will just be ignored by `fog`.

I've asked fog guys when they are going to make a release [here](https://github.com/fog/fog/pull/3665#issuecomment-132313339)

PR #1725, opened 2015-08-23
Fix Multiple file uploads section in README.md

rusikf:
I think, there is must be more description about  uploading multiple files on different sql adapters.


bensie:
Closing in favor of https://github.com/carrierwaveuploader/carrierwave/issues/1755...you're right -- we need better docs for that.

PR #1731, opened 2015-09-04
Remove rails-master gemfile from travis checks

eavgerinos:
rails/rails@master is currently being developed for the Rails 5 version.
At the moment, the actionpack-5.0.0.alpha depends on rack-2.0.0.alpha
which is not published in rubygems.org on in any of rack's branches,
thus leading to failures in our build.


bensie:
@eavgerinos possible to allow failures on it instead of deleting it?

eavgerinos:
@bensie You are right. Silly me :)

PR #1738, opened 2015-09-10
README change. added helping notes around paper_trail and multiple file uploads issue

sjahandideh:
Finding the workaround took me a while, so I added this little description to README file. this only happens with multiple file uploads. the reason is YAML.dump can't dump Uploader images. YAML.dump is used inside paper_trail default serializer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1006,6 +1006,18 @@ class User
 end
 ```
 
+## Using with PaperTrail
+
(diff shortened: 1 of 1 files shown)


bensie:
Thanks for this, but it's a bit too specific to another project. Feel free to add this to the wiki though!

PR #1739, opened 2015-09-11
Updated Readme with note about array column

bigzed:
I found it cumbersome to find get MySql to work with multiple file uploads. The issue was in the end that I forgot to add `serialize :avatars` to my model. You may want to include this in your README.

best,
Steve

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,12 @@ class User < ActiveRecord::Base
 end
 ```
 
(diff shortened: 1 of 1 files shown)


bensie:
This may work, but I don't think it should be documented as officially supported. It's currently targeting PostgreSQL and should work with MySQL 5.7, which has native JSON columns.

PR #1746, opened 2015-09-24
Fix `Mounter.blank?` method

Bonias:


bensie:
@Bonias Can you explain this change?

Bonias:
I just wanted to point out that `blank?` method is currently broken and tests don't cover it, but I'm not sure that my changes are proper way to fix it or test it.

How it is working now:

    @instance.image = nil
    @instance.image? # => true

What I expect:

    @instance.image = nil
    @instance.image? # => false

Looks like problem was introduced here https://github.com/carrierwaveuploader/carrierwave/commit/e38789cd40a9a94ab5e281a98e6e83da26b27036#diff-bff15cc47155b179e2ad618ed40ebbe2L92. Earlier `uploader.blank?` was called, which was defined as `file.blank?`, so we were checking if uploader had file assigned. At current implementation `uploaders.empty?` (`Array#empty?`) method is used. This mean that `true` (wrongly) would be returned when `uploaders` includes only one uploader without a file.

--

Why current tests does not cover it?

When `expect(@instance).to receive(:read_uploader).with(:image).and_return('')` is used 
`read_identifiers` (https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/mounter.rb#L27) returns empty array and thus `uploaders` (https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/mounter.rb#L31) returns empty array.

When we use `@instance.image=` then `cache` method (https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/mounter.rb#L39) is used which sets `@uploaders` and now `@uploaders` consist of one uploader without a file which makes whole that truble.

thomasfedb:
Is the return value of `blank?` defined anywhere in documentation or comments?

thomasfedb:
@bensie This appears like a reasonable change with good justification provided by @Bonias, can it be merged?

bensie:
Thanks @Bonias!

PR #1753, opened 2015-10-09
Get rid of SSRF

anthony07:
Error messages with the remote download is pretty open and I think it's a big vulnerability issue.

### DIFF
--- a/lib/carrierwave/uploader/download.rb
+++ b/lib/carrierwave/uploader/download.rb
@@ -43,8 +43,8 @@ module CarrierWave
           @file
 
         rescue StandardError => e
-          reason = e.is_a?(Net::HTTPBadResponse) || e.is_a?(Errno::EHOSTUNREACH) ? '' : ": #{e.message}"
-          raise CarrierWave::DownloadError, "could not download file#{reason}"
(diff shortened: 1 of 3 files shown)


lib/carrierwave/uploader/download.rb
```diff
@@ -43,7 +43,8 @@ def file
           @file
 
         rescue StandardError => e
-          raise CarrierWave::DownloadError, "could not download file: #{e.message}"
+          reason = e.is_a?(Net::HTTPBadResponse) || e.is_a?(Errno::EHOSTUNREACH) ? '' : ": #{e.message}"
```
bensie: Can you explain this further? Why are messages swallowed for these two exception classes?

thomasfedb:
@anthony07 Could you provide some more context on the situations in which this change is helpful? Some examples of what was wrong the the old behaviour would be good.

lib/carrierwave/version.rb
```diff
@@ -1,3 +1,3 @@
 module CarrierWave
-  VERSION = "0.10.0"
+  VERSION = "0.10.2"
```
thomasfedb: We can't merge a PR that changes `version.rb`.

thomasfedb:
@anthony07 I'm closing this PR as I've not heard back from you. Let me know if you'd like it reopened.

PR #1758, opened 2015-10-22
Add be_format test matcher

yanivpr:
Add a test matcher for the `format`.
Could be useful to test that a conversion has happened.

Usage:
```
context 'the thumb format' do
  it 'should be JPEG' do
    @uploader.thumb.should be_format('JPEG')
   end
end
```

### DIFF
--- a/spec/processing/mini_magick_spec.rb
+++ b/spec/processing/mini_magick_spec.rb
@@ -23,9 +23,8 @@ describe CarrierWave::MiniMagick do
   describe "#convert" do
     it "should convert from one format to another" do
       @instance.convert('png')
-      img = ::MiniMagick::Image.open(@instance.current_path)
-      expect(img['format']).to match(/PNG/)
       expect(@instance.file.extension).to eq('png')
+      expect(@instance).to be_format('png')
     end
 
     it "should convert all pages when no page number is specified" do
--- a/spec/processing/rmagick_spec.rb
+++ b/spec/processing/rmagick_spec.rb
@@ -22,8 +22,8 @@ describe CarrierWave::RMagick, :rmagick => true do
   describe '#convert' do
     it "should convert the image to the given format" do
       @instance.convert(:png)
-      expect(::Magick::Image.read(@instance.current_path).first.format).to eq('PNG')
       expect(@instance.file.extension).to eq('png')
+      expect(@instance).to be_format('png')
     end
   end
 
--- a/README.md
+++ b/README.md
@@ -590,6 +590,10 @@ describe MyUploader do
   it "should make the image readable only to the owner and not executable" do
     @uploader.should have_permissions(0600)
   end
+
+  it "should be the correct format" do
+    @uploader.should be_format('png')
+  end
 end
 ```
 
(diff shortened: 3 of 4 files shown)


thomasfedb:
@yanivpr does the value of `@expected` need to be normalised at all? e.g. `@expected.downcase`?

yanivpr:
@thomasfedb To clarify - you mean that I should cover all cases?
Something like `/\A#{@actual_expected}\z/i.match(@expected)`

thomasfedb:
@yanivpr is a regex the most efficient way to handle that?

thomasfedb:
Also @yanivpr, it might be worthwhile ensuring that your matcher gets used within the carrierwave test suite.

yanivpr:
* Test format case insensitive.
* Use format test matcher in the test suite.
* Add format test matcher to readme file.

spec/processing/mini_magick_spec.rb
```diff
@@ -23,9 +23,7 @@
   describe "#convert" do
     it "should convert from one format to another" do
       @instance.convert('png')
-      img = ::MiniMagick::Image.open(@instance.current_path)
-      expect(img['format']).to match(/PNG/)
-      expect(@instance.file.extension).to eq('png')
+      expect(@instance).to be_format('PNG')
```
thomasfedb: Can you change `'PNG'` to `'png'`.

spec/processing/mini_magick_spec.rb
```diff
@@ -23,9 +23,7 @@
   describe "#convert" do
     it "should convert from one format to another" do
       @instance.convert('png')
-      img = ::MiniMagick::Image.open(@instance.current_path)
-      expect(img['format']).to match(/PNG/)
-      expect(@instance.file.extension).to eq('png')
```
thomasfedb: I think we need to include the file extension check here, unless you'd also like to write a file extension matcher. :smile:

spec/processing/rmagick_spec.rb
```diff
@@ -22,8 +22,7 @@
   describe '#convert' do
     it "should convert the image to the given format" do
       @instance.convert(:png)
-      expect(::Magick::Image.read(@instance.current_path).first.format).to eq('PNG')
-      expect(@instance.file.extension).to eq('png')
```
thomasfedb: And here also.

thomasfedb:
This is looking great @yanivpr. If you could address the three comments I've left above, and also squash all your changes into a single commit ([how to](http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git)) then I'll git it merged! :trophy:

spec/processing/rmagick_spec.rb
yanivpr: :+1:

spec/processing/mini_magick_spec.rb
yanivpr: :+1:

spec/processing/mini_magick_spec.rb
yanivpr: :+1: Putting back the `file.extension` expectation without a new matcher :)

yanivpr:
Thank you for the code review!
I have squashed the commits.

thomasfedb:
Lovely. Cheers @yanivpr.

PR #1763, opened 2015-10-31
Fix Get Started documentation

abelorian:
little mistake


thomasfedb:
@bensie can you verify and merge this please.

bensie:
@abelorian Did you verify that `avatar_identifier` doesn't work? It should be correct as-is...

thomasfedb:
@bensie is correct, the mounter API includes the `#{column}_identifier` method.

https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/mount.rb#L182

PR #1766, opened 2015-11-04
Retrieve old file using identifier string not changes array itself

nanocity:
Hi all,

I use carrierwave with [Cloudinary uploader](https://github.com/cloudinary/cloudinary_gem) and I'm randomly seeing this error.

    undefined method `match' for ["image/upload/v111111/aaaaaaaaaa.png"]:Array
    vendor/bundle/ruby/2.2.0/gems/cloudinary-1.1.0/lib/cloudinary/carrier_wave.rb:150 in initialize
    vendor/bundle/ruby/2.2.0/gems/cloudinary-1.1.0/lib/cloudinary/carrier_wave.rb:29 in new
    vendor/bundle/ruby/2.2.0/gems/cloudinary-1.1.0/lib/cloudinary/carrier_wave.rb:29 in retrieve_from_store!
    vendor/bundle/ruby/2.2.0/bundler/gems/carrierwave-49fdad1ec6ca/lib/carrierwave/mounter.rb:130 in block in remove_previous
    vendor/bundle/ruby/2.2.0/bundler/gems/carrierwave-49fdad1ec6ca/lib/carrierwave/mounter.rb:127 in map
    vendor/bundle/ruby/2.2.0/bundler/gems/carrierwave-49fdad1ec6ca/lib/carrierwave/mounter.rb:127 in remove_previous
    vendor/bundle/ruby/2.2.0/bundler/gems/carrierwave-49fdad1ec6ca/lib/carrierwave/mount.rb:192 in remove_previously_stored_photo
    vendor/bundle/ruby/2.2.0/gems/activesupport-4.2.3/lib/active_support/callbacks.rb:430 in block in make_lambda
    vendor/bundle/ruby/2.2.0/gems/activesupport-4.2.3/lib/active_support/callbacks.rb:251 in call
    vendor/bundle/ruby/2.2.0/gems/activesupport-4.2.3/lib/active_support/callbacks.rb:251 in block in conditional
    ...

I think CarrierWave is passing an array to retrieve_from_store! method on uploader when it should pass a plain string.

    before = before.reject(&:blank?).map do |value|
        if value.is_a?(String)
          uploader = blank_uploader
          uploader.retrieve_from_store!(before)
          uploader
        else
          value
        end
      end

Tests are green, probably because ["file"]:Array and "file":String are equivalent in many cases but Clodinary uploader expect a String to call match method.

I really don't know why my app is not failing consistently, but this fix doesn't break tests and code make me think there is a typo. Also the parameter for retrieve_from_store is always a String so clodinary error should disappear now.

Sorry for my english.

### DIFF
--- a/lib/carrierwave/mounter.rb
+++ b/lib/carrierwave/mounter.rb
@@ -127,7 +127,7 @@ module CarrierWave
       before = before.reject(&:blank?).map do |value|
         if value.is_a?(String)
           uploader = blank_uploader
-          uploader.retrieve_from_store!(before)
+          uploader.retrieve_from_store!(value)
           uploader
         else
           value


thomasfedb:
Looks like a sensible change, although I'm not familiar with this part of the codebase. @nanocity are you able to provide an example that reproduces the error you've been seeing, or better yet, a failing test?

nanocity:
Sorry, I can't provide a failing test because CarrierWave code is not broken.

The thing is, CarrierWave is sending an Array to retrieve_from_store! method but third party library expects a String. If we only consider this, I should send a PR to Clodinary Gem and let them fix the issue. The problem is that source code of remove_previous method makes me think that this behavior is not part of a specification but a typo.

thomasfedb:
@nanocity I agree, it looks like a typo. Also, I'm certain that a test can be written to ensure that Carrierwave performs lookups using a string rather than an array.

thomasfedb:
@QuotableWater7 You wrote the code in question, any input?

QuotableWater7:
@thomasfedb  I didn't actually change the behavior of the code if you look at my change - I just un-nested the code in the method.  the previous changes might be more useful to look at to see who the author is

thomasfedb:
Right you are @QuotableWater7. @jnicklas, looks like you introduced this in 62eee42, care to offer your thoughts?

jnicklas:
@thomasfedb I'm pretty sure it's a bug and your PR is correct.

thomasfedb:
Just going to run this build again before I look to merge it.

PR #1772, opened 2015-11-16
Allow non-ASCII filename by default. Rework of #1148

shuhei:
BREAKING CHANGE: Allowing non-ASCII filename by default.

This is a rework of #1148 by @yalab.

### DIFF
--- a/lib/carrierwave/sanitized_file.rb
+++ b/lib/carrierwave/sanitized_file.rb
@@ -28,7 +28,7 @@ module CarrierWave
       attr_writer :sanitize_regexp
 
       def sanitize_regexp
-        @sanitize_regexp ||= /[^a-zA-Z0-9\.\-\+_]/
+        @sanitize_regexp ||= /[^[:word:]\.\-\+]/
       end
     end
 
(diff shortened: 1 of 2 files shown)


thomasfedb:
LGTM :sparkles: 

@bensie Is it possible to either get master passing on Travis, or determine if CI failures in this build are related to the changes in this PR.

bensie:
@thomasfedb I think there was a problem with Travis when this ran initially, re-running seems to be working fine. There is an issue with the `rails-master` Gemfile (Rails 5), but this PR isn't affecting that.

This looks great, thanks @shuhei!

shuhei:
@bensie @thomasfedb Thanks for reviewing and merging!

shuhei:
@bensie Is this gonna be released in the near future?

thomasfedb:
@shuhei This will be included in the 1.0.0 release, which there isn't a defined timeline for quite yet.

shuhei:
@thomasfedb I understand that this should be in a major upgrade because it breaks the current default. Thanks for all your effort for keeping this great library working!

PR #1787, opened 2015-11-30
Allow failures in CI builds against Rails master

thomasfedb:
CI builds against Rails master are currently failing as Rails 5 depends on an unpublished version of Rack.


thomasfedb:
@bensie Could you please merge this at your convenience.

bensie:
The version of JRuby we're testing against doesn't support keyword args, so something that was recently added is now causing that to fail :(

Issue #1232, opened 2013-10-02
Possible generate_cache_id clash


ronalchn:
The `generate_cache_id` method generates the directory in which a temporary file is placed.

```
def self.generate_cache_id
  Time.now.utc.to_i.to_s + '-' + Process.pid.to_s + '-' + ("%04d" % rand(9999))
end
```

This could clash easily if the same filename is uploaded within the same minute. This could occur if two users try to upload the same filename at the same time, especially for single worker servers (only 1 pid). It may also occur if a user uploads the same filename to two different fields in the same form at once.

The random term at the end doesn't eliminate the possibility, it only makes it less likely, but there is still a 0.01% chance. The chance of hash collisions are usually considered almost impossible, but the chance of that is often on the order of in it happening once in a billion years, even if used a million times a day.

I propose that the random term is changed to a counter term (separate counter variables for each process). A counter makes it impossible for it to happen unless there are enough requests for a single process working fast enough to cause it to wrap within the same minute. Because of the possibility of multi-threaded, single process servers, an extra thread id term should also be added.

PR #1797, opened 2015-12-14
Add counter to cache_id

stillwaiting:
Current (in master) version of generate_cache_id is collision-prone: for complex factories with multiple
uploaders we have quite a few flaky specs because of that. Having a counter, this and similar situations, like mentioned [here](https://github.com/carrierwaveuploader/carrierwave/issues/1232), should be resolved completely.

### DIFF
--- a/lib/carrierwave/uploader/cache.rb
+++ b/lib/carrierwave/uploader/cache.rb
@@ -7,16 +7,28 @@ module CarrierWave
       "You tried to assign a String or a Pathname to an uploader, for security reasons, this is not allowed.\n\n If this is a file upload, please check that your upload form is multipart encoded."
     end
   end
+  
+  class CacheCounter
+    @@counter = 0
+
+    def self.increment
+      @@counter += 1
+    end  
+  end
 
   ##
   # Generates a unique cache id for use in the caching system
   #
   # === Returns
   #
-  # [String] a cache id in the format TIMEINT-PID-RND
+  # [String] a cache id in the format TIMEINT-PID-COUNTER-RND
   #
   def self.generate_cache_id
-    Time.now.utc.to_i.to_s + '-' + Process.pid.to_s + '-' + ("%04d" % rand(9999))
+    [Time.now.utc.to_i,
+      Process.pid, 
+      '%04d' % (CarrierWave::CacheCounter.increment % 1000), 
+      '%04d' % rand(9999)
+    ].map(&:to_s).join('-')
   end
 
   module Uploader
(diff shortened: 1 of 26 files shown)


lib/carrierwave/uploader/cache.rb
```diff
@@ -178,7 +189,9 @@ def workfile_path(for_file=original_filename)
       alias_method :full_original_filename, :original_filename
 
       def cache_id=(cache_id)
-        raise CarrierWave::InvalidParameter, "invalid cache id" unless cache_id =~ /\A[\d]+\-[\d]+\-[\d]{4}\z/
+        # Earlier cache_id contained only 3 pieces, the "counter" part was introduced later.
```
thomasfedb: No need to talk about `later`. Change to "Earlier version used 3 part cache_id."

lib/carrierwave/uploader/cache.rb
```diff
@@ -7,16 +7,27 @@ def message
       "You tried to assign a String or a Pathname to an uploader, for security reasons, this is not allowed.\n\n If this is a file upload, please check that your upload form is multipart encoded."
     end
   end
+  
+  class CacheCounter
+    @@counter = 0
+    def self.increment
```
thomasfedb: Please insert a linebreak before this method definition.

thomasfedb:
@bensie can I get your thoughts on this?

thomasfedb:
CI is failing on JRuby and Rails master, both of which are expected/

stillwaiting:
Seems they are failing for quite a while, e.g. here I can see the same picture: https://travis-ci.org/carrierwaveuploader/carrierwave/builds/96401284

thomasfedb:
@stillwaiting yes, that was what I mean, that those cases also fail on `master` and the issue is not introduced by your changes.

thomasfedb:
GTTM @bensie

mehlah:
@thomasfedb @stillwaiting The build should be :green_heart: now thanks to #1807.
I reviewed the changes, and it looks good to me :+1:

mtsmfm:
@thomasfedb How about backporting?
Our tests will fail if collision occurs :cry: 
Many carrierwave users may tackle this random failure.

thomasfedb:
@mtsmfm I'm very happy to consider back porting this to the 0.10-stable branch, although we need to get the CI build fixed first.

mtsmfm:
OK, I'll investigate that :wink:

mehlah:
@mtsmfm should be good with #1847 cc @thomasfedb.
We can start backporting some fixes for an intermediate release.

mtsmfm:
@mehlah Thanks for letting me know :smile:

PR #1803, opened 2015-12-29
add a process command: resize_to_extend

hotvulcan:
I add a process command: resize_to_extend for mini magick processings. It resize the image to fill target dimension and chop the extend part(s).  This process works well in my project.


thomasfedb:
@hotvulcan, can you explain how this differs from `resize_to_fill` which is documented as follows:

> Resize the image to fit within the specified dimensions while retaining
> the aspect ratio of the original image. If necessary, crop the image in the
> larger dimension.

thomasfedb:
If you feel that this offers something over what's provided by `resize_to_fill` then I will reopen this PR for discussion. Otherwise, however, I'm closing this PR.

PR #1804, opened 2015-12-29
Make `CarrierWave::MiniMagick#mini_magic_image` private

mehlah:
I think `CarrierWave::MiniMagick#mini_magic_image` method doesn't need to be part of the public API, at least to keep a symmetry between MiniMagick and RMagick processors (more at #1805). 
This method is used by the new public `#width` and `#height` methods (introduced by #1405).
This commit also fixes a typo in the method name (magick instead of magic).

Any thoughts?

### DIFF
--- a/lib/carrierwave/processing/mini_magick.rb
+++ b/lib/carrierwave/processing/mini_magick.rb
@@ -245,40 +245,25 @@ module CarrierWave
     end
 
     ##
-    # Return the mini magic instance of the image
-    #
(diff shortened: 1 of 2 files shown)


thomasfedb:
Thanks @mehlah. I'm happy with this change. While it represents a breaking change, we're going into a major version release, so that's acceptable.

I'm going to restart the build on CI, and if it passes then I'll get this merged.

PR #1805, opened 2015-12-29
Add `#width` and `#height` methods to the RMagick processor

mehlah:
This keeps the symmetry with MiniMagick processor where `#width` and
`#height` methods has been introduced by #1405.

### DIFF
--- a/lib/carrierwave/processing/rmagick.rb
+++ b/lib/carrierwave/processing/rmagick.rb
@@ -252,6 +252,28 @@ module CarrierWave
     end
 
     ##
+    # Returns the width of the image.
+    #
+    # === Returns
+    #
+    # [Integer] the image's width in pixels
+    #
+    def width
+      rmagick_image.columns
+    end
+
+    ##
+    # Returns the height of the image.
+    #
+    # === Returns
+    #
+    # [Integer] the image's height in pixels
+    #
+    def height
+      rmagick_image.rows
+    end
+
+    ##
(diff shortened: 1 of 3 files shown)


lib/carrierwave/processing/rmagick.rb
```diff
@@ -351,5 +373,9 @@ def destroy_image(image)
       image.destroy! if image.respond_to?(:destroy!)
     end
 
+    def rmagick_image
+      ::Magick::Image.ping(current_path).first
```
mehlah: I used `ping` instead of `read` as we are interested in the image metadata, but don't care about the pixel data.
Ref http://www.rubydoc.info/gems/rmagick/2.15.4/Magick/Image#ping-class_method

thomasfedb: I'm a little concerned about this. Calling `ping` will return an image with invalid pixel data, and naming this `rmagick_image` seems to be asking for trouble.

Either we need to read the entire image, or we should rename the method.

thomasfedb:
I'm happy with this change in concept, but I've noted one concern above.

mehlah:
@thomasfedb I agree :+1:. I fixed it by using `read` instead of `ping` to keep it consistent with MiniMagick code.
I also added an additional commit to remove an unnecessary `private` keyword introduced after recent merges (see the commit message for details).
Thanks!

thomasfedb:
Thanks @mehlah!

PR #1754, opened 2015-10-12
process mini magic commands with combine_options

bernabas:

PR #1806, opened 2015-12-29
Make `CarrierWave::MiniMagick#append_combine_options` private

mehlah:
This method isn't intended to be part of the public API.
It's used internally by the processor to accept `combine_options`
introduced at #1754.


thomasfedb:
LGTM. Thanks @mehlah. Restarting build and will merge if that passes.

PR #1812, opened 2016-01-01
Use `try` instead of `obj.foo if obj.respond_to?(:foo)`

mehlah:
As CarrierWave already depends on ActiveSupport that brings `try`,
this change helps to improve the code readability avoiding conditionals.

### DIFF
--- a/lib/carrierwave/sanitized_file.rb
+++ b/lib/carrierwave/sanitized_file.rb
@@ -164,9 +164,9 @@ module CarrierWave
       elsif is_path?
         File.open(@file, "rb") {|file| file.read}
       else
-        @file.rewind if @file.respond_to?(:rewind)
+        @file.try(:rewind)
         @content = @file.read
-        @file.close if @file.respond_to?(:close) && @file.respond_to?(:closed?) && !@file.closed?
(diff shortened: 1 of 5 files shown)


lib/carrierwave/sanitized_file.rb
```diff
@@ -164,9 +164,9 @@ def read
       elsif is_path?
         File.open(@file, "rb") {|file| file.read}
       else
-        @file.rewind if @file.respond_to?(:rewind)
+        @file.try(:rewind)
         @content = @file.read
-        @file.close if @file.respond_to?(:close) && @file.respond_to?(:closed?) && !@file.closed?
+        @file.try(:close) if !@file.try(:closed?)
```
thomasfedb: Let change `if !@file.try(:closed?)` to `unless @file.try(:closed?)`.

thomasfedb:
LGTM, other than the comment above.

lib/carrierwave/sanitized_file.rb
mehlah: :+1:

mehlah:
Comment addressed :+1:

PR #1813, opened 2016-01-01
Remove `CarrierWave::MimeTypes` processor module

mehlah:
Since #1245 (*), CarrierWave::MimeTypes module is redundant as `mime-types` gem is a runtime dependency now, and `SanitizedFile` gets it's content_type using it.
The module was deprecated since 0.10 and warned developers when used it.

What do you think about removing MagicMimeTypes processor module too?
I can't see any valid reason to keep it around.
A related side note: CarrierWave rely on ruby-filemagic gem (`MagicMime`) to validates the content type of a file against a white/blacklist.
I think we should use mime-types (`Mime::Types`) for this as it's a runtime dependency, and it's consistent with what it's used by `SanitizedFile` to set a content type.
Otherwise, we could use `MagicMime` everywhere as a runtime dependency instead of `Mime::Types`, but this kills JRuby support.

(*) I discovered this when tried to fix specs marked as pending. I looked at commits history and PRs to understand what happened.
It started with #372 that added `CarrierWave::MimeTypes`  processor (Jun 2011), then #1216 and #1245 that made `mime-types` a runtime dependency and deprecated `CarrierWave::MimeTypes` module (2013), and finally #1584 `CarrierWave::MagicMimeTypes` (2015)

PR #1816, opened 2016-01-02
Remove ruby-filemagic dependency and refactor/redesign content type white/blacklisting feature

mehlah:
This PR removes `CarrierWave::MagicMimeTypes` processor module at d6fa112 (see #1813 for background).

As discussed with @thomasfedb (https://github.com/carrierwaveuploader/carrierwave/pull/1813#issuecomment-168316910), it also remove the dependency on `ruby-filemagic` in white/blacklist content types (b5bec3d).

Background: Before this, CarrierWave was relying on mime-types or ruby-filemagic
gems to get/set the content type of a file in an inconsistent way.
After making mime-types a runtime dependency for CW,
`CarrierWave::MimeTypes` has been dropped, and `CarrierWave::MagicMimeTypes` as well with this.

White/blacklisting content types feature (introduced by #1596) was relying only on ruby-filemagic to read the content type of a file.
It was optional and required mixing a module to the uploader.

This PR removes this dependency by using the already available content type (thanks to `Mime::Types`).
This let us drop any dependency on ruby-filemagic in CarrierWave.
It makes JRuby support better as ruby-filemagic is unsupported on JRuby due to the C extension.

I also made some refactors, renamed methods like `whitelist_mime_type_pattern` to `content_type_whitelist`, cleaned up specs and docs.
The module doesn't need to be optional now, so it's included by default (like `ExtensionWhitelist` and `ExtensionBlacklist`)

To keep it consistent with the current way of white/blacklisting extensions, `content_type_whitelist` now accepts an array of strings or regexps to white/blacklist content types. I can move 07c80cb to a separate PR if you prefer.


spec/uploader/callback_spec.rb
```diff
@@ -3,19 +3,27 @@
 describe CarrierWave::Uploader do
 
   it "should keep callbacks on different classes isolated" do
+    default_before_callbacks = [
+      :check_whitelist!,
```
thomasfedb: Might be worth changing `check_whitelist!` to `check_extension_whitelist!` and similar. Happy to do it as a separate PR though.

mehlah: I actually did it already :bowtie:
https://github.com/mehlah/carrierwave/compare/remove-magicmime...mehlah:refactor-extensions-white/blacklist#diff-2bca1d8b1d06f3d29a6553468c7b940fR7
I'll PR the refactor for better consistency as soon as this is accepted.

thomasfedb: :+1:

mehlah:
I screwed up 6df8c8341be7c87870dedd328e74b88f7a7268a9 after a rebase (made the build red https://travis-ci.org/carrierwaveuploader/carrierwave/jobs/99844432). Now fixed ;)

PR #1817, opened 2016-01-02
Avoid duplication between RSpec and Cucumber AR support config

mehlah:
Since `spec/support/activerecord.rb` is required in `features/support/activerecord.rb`, duplicated AR config could be avoided (only migrations differs).
`raise_in_transactional_callbacks` should/would be set to `true` not only for Rails 4.2 but >= 4.2.
Related to #1815.


thomasfedb:
Thanks @mehlah

PR #1824, opened 2016-01-03
Refactor AR specs

mehlah:
It all started when I read @jnicklas comment:
```ruby
# My god, what a horrible, horrible solution, but AR validations don't work
# unless the class has a name. This is the best I could come up with :S
```
I tried to make it better :bowtie:
This PR mainly refactor the way AR classes are reset in specs (e18499569792341d7f10665e226a9c1ca3fe0305).
It uses `Object#remove_const` and `Object#const_set` instead of initializing a new `Event{$counter}` before each spec example.
I think we could extract this support code to a separate module to make it even cleaner.

It also uses ShamRack only around a couple of specs examples that needs it, and not for each one.

I tried to reduce nested `before` callbacks in the examples concerned by this change to make those easier to read.

Minor changes:
- Remove unneeded AR config to not include root in JSON
- Replace AR migrations in specs by `ActiveRecord::Base.connection` calls

### DIFF
--- a/spec/orm/activerecord_spec.rb
+++ b/spec/orm/activerecord_spec.rb
@@ -25,8 +25,6 @@ describe CarrierWave::ActiveRecord do
   after(:all) { TestMigration.down }
 
   before do
-    # Rails 4 defaults to no root in JSON, join the party
-    ActiveRecord::Base.include_root_in_json = false
     # My god, what a horrible, horrible solution, but AR validations don't work
     # unless the class has a name. This is the best I could come up with :S
     $arclass += 1
--- a/spec/orm/activerecord_spec.rb
+++ b/spec/orm/activerecord_spec.rb
@@ -17,27 +17,19 @@ class TestMigration < ActiveRecord::Migration
   end
 end
 
-class Event < ActiveRecord::Base; end # setup a basic AR class for testing
-$arclass = 0
+def reset_class(class_name)
+  Object.send(:remove_const, class_name) rescue nil
+  Object.const_set(class_name, Class.new(ActiveRecord::Base))
+end
 
 describe CarrierWave::ActiveRecord do
   before(:all) { TestMigration.up }
   after(:all) { TestMigration.down }
 
   before do
-    # My god, what a horrible, horrible solution, but AR validations don't work
-    # unless the class has a name. This is the best I could come up with :S
-    $arclass += 1
-
-    @class = Class.new(Event)
-    # AR validations don't work unless the class has a name, and
-    # anonymous classes can be named by assigning them to a constant
-    Object.const_set("Event#{$arclass}", @class)
-    @class.table_name = "events"
     @uploader = Class.new(CarrierWave::Uploader::Base)
-    @class.mount_uploader(:image, @uploader)
-    @class.mount_uploaders(:images, @uploader)
-    @event = @class.new
+    reset_class("Event")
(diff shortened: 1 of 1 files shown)


thomasfedb:
I actually have already created a gem which is specifically for creating AR models for tests. https://github.com/thomasfedb/jamjar

thomasfedb:
Just a note @mehlah, I'm heading out of town for a week tomorrow, and I'll only have internet access occasionally. I'll try to keep up to date with CarrierWave PRs, but forgive me if I get behind.

spec/orm/activerecord_spec.rb
```diff
@@ -1195,6 +1197,14 @@ def monkey
     end
 
     describe "#remote_images_urls=" do
+      before do
+        sham_rack_app = ShamRack.at('www.example.com').stub
+        sham_rack_app.register_resource('/test.jpg', File.read(file_path('test.jpg')), 'images/jpg')
```
thomasfedb: I'd much rather extract this into a support module which provides a macro. Perhaps `setup_remote_url(url, file, mime)`, or perhaps `mime` can be derived from `file` automatically.

thomasfedb:
I'd also like to see instance variables removed from this file, and replace with `let(:var) { value }`.

spec/orm/activerecord_spec.rb
mehlah: :+1:

thomasfedb:
@mehlah Do you intend to make any further changes in this PR?

mehlah:
@thomasfedb sure! I'll work on it probably on Friday.

spec/orm/activerecord_spec.rb
mehlah: @thomasfedb I started extracting a support module for ShamRack but felt this wasn't the best and simplest way to go (espacially when looking at other ShamRack stubs in `download_spec` for example).
I suggest replacing ShamRack by WebMock which provides imho a better API for stubbing external HTTP requests and it's widely used.
I replaced ShamRack by WebMock at https://github.com/mehlah/carrierwave/commit/c28633e13eb83bc49c544bf9835966fbb5aef048 to simplify/clarify some specs examples.
I'd be happy to PR this if you want to discuss it.

mehlah:
> I'd also like to see instance variables removed from this file, and replaced with `let(:var) { value }`.

`let` or `let!` doesn't seems to be used by other CW specs.
I think it would be better to keep instances variables to be consistent with the whole test suite.

thomasfedb:
CarriewWave specs in general use a fairly outdated RSpec style that's I'd like to see progressively modernised.

spec/orm/activerecord_spec.rb
thomasfedb: Yeah, a PR would be great. Much easier for me to look over on my mobile (out of town at the moment).

mehlah: Opened at #1829 :+1: 
I'll rebase this one depending on which direction we are going.

thomasfedb:
Hey @mehlah, could you rebase this please?

mehlah:
@thomasfedb I rebased this against master. Should be ready for a merge.
I suggest to address replacing ivars in specs by `let` or  `let!` in a separate PR for all the specs. Good for you?

thomasfedb:
Thanks @mehlah! :)

PR #1825, opened 2016-01-05
Make extensions and content types white/blacklists take a string, regexp, or array of values

mehlah:
Writing

```ruby
def content_type_whitelist
  [/image\//]
end
```
isn't very convenient or intuitive due to the array.

This PR suggests to let `#content_type_whitelist` and `extension_whitelist` takes either a
string, a regexp, or an array of values (same thing for blacklists).

### DIFF
--- a/lib/carrierwave/uploader/content_type_blacklist.rb
+++ b/lib/carrierwave/uploader/content_type_blacklist.rb
@@ -14,7 +14,7 @@ module CarrierWave
       #
       # === Returns
       #
-      # [NilClass, Array[String,Regexp]] a blacklist of content types which are not allowed to be uploaded
+      # [NilClass, String, Regexp, Array[String, Regexp]] a blacklist of content types which are not allowed to be uploaded
       #
       # === Examples
       #
@@ -40,7 +40,7 @@ module CarrierWave
       end
 
       def blacklisted_content_type?(content_type)
-        content_type_blacklist.any? { |item| content_type =~ /#{item}/ }
+        Array(content_type_blacklist).any? { |item| content_type =~ /#{item}/ }
       end
 
     end # ContentTypeBlacklist
--- a/lib/carrierwave/uploader/content_type_whitelist.rb
+++ b/lib/carrierwave/uploader/content_type_whitelist.rb
@@ -14,7 +14,7 @@ module CarrierWave
       #
       # === Returns
       #
-      # [NilClass, Array[String,Regexp]] a whitelist of content types which are allowed to be uploaded
+      # [NilClass, String, Regexp, Array[String, Regexp]] a whitelist of content types which are allowed to be uploaded
       #
       # === Examples
       #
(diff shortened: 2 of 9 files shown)


lib/carrierwave/uploader/content_type_blacklist.rb
```diff
@@ -14,7 +14,7 @@ module ContentTypeBlacklist
       #
       # === Returns
       #
-      # [NilClass, Array[String,Regexp]] a blacklist of content types which are not allowed to be uploaded
+      # [NilClass, String, Regexp, Array[String,Regexp]] a blacklist of content types which are not allowed to be uploaded
```
thomasfedb: For consistency, can you add a space in `Array[String,Regexp]` so that it becomes `Array[String, Regexp]`.

lib/carrierwave/uploader/content_type_whitelist.rb
```diff
@@ -14,7 +14,7 @@ module ContentTypeWhitelist
       #
       # === Returns
       #
-      # [NilClass, Array[String,Regexp]] a whitelist of content types which are allowed to be uploaded
+      # [NilClass, String, Regexp, Array[String,Regexp]] a whitelist of content types which are allowed to be uploaded
```
thomasfedb: And here also.

lib/carrierwave/uploader/extension_blacklist.rb
```diff
@@ -17,7 +17,7 @@ module ExtensionBlacklist
       #
       # === Returns
 
-      # [NilClass, Array[String,Regexp]] a black list of extensions which are prohibited to be uploaded
+      # [NilClass, String, Regexp, Array[String,Regexp]] a black list of extensions which are prohibited to be uploaded
```
thomasfedb: Here also.

lib/carrierwave/uploader/extension_whitelist.rb
```diff
@@ -17,7 +17,7 @@ module ExtensionWhitelist
       #
       # === Returns
       #
-      # [NilClass, Array[String,Regexp]] a white list of extensions which are allowed to be uploaded
+      # [NilClass, String, Regexp, Array[String,Regexp]] a white list of extensions which are allowed to be uploaded
```
thomasfedb: And here.

thomasfedb:
Can you rename the relevant specs to be in keeping with the names for your new specs, eg: `"accepts a single extension string value"` would have the counterpart `"accepts an array of extension string values"`.

Or even better, something like:

```ruby
context "when there is a blacklist" do
  context "when the backlist is a single value" do
    # ...
  end

  context "when the backlist is an array of values" do
    # ...
  end
end
```

mehlah:
@thomasfedb Thanks for your comments. I addressed those to make the specs perfectly symmetric and rebased against master.

thomasfedb:
Thanks @mehlah! :sparkles:

PR #1827, opened 2016-01-08
Cleanup the gemspec

mehlah:
This removes from the gemspec

- `rubyforge_project` as it's no more a thing
- `date`, `rubygems_version` and `specification_version` as those shouldn't be set as recommended.

See:
http://www.rubydoc.info/github/rubygems/rubygems/Gem%2FSpecification%3Aspecification_version
http://www.rubydoc.info/github/rubygems/rubygems/Gem/Specification#rubygems_version-instance_method
http://www.rubydoc.info/github/rubygems/rubygems/Gem/Specification#date%3D-instance_method


thomasfedb:
Thanks @mehlah!

PR #1828, opened 2016-01-09
URI.decode (URI.unescape) is obsolete

koic:
[URI.decode is obsolete](https://github.com/ruby/ruby/blob/ruby_2_3/lib/uri/common.rb#L125).
And `CarrierWave::Storage::Fog::File#filename` method is returning a string with ASCII-8BIT encoding on Ruby 2.3.0.

The problem is that it can't decode a multi-byte characters containing URL.

Thanks.

### DIFF
--- a/spec/storage/fog_spec.rb
+++ b/spec/storage/fog_spec.rb
@@ -36,5 +36,16 @@ describe CarrierWave::Storage::Fog::File do
         expect(subject.filename).to eq('foo.txt')
       end
     end
+
+    context "when url contains multi-byte characters ('日本語') in query string" do
+      before do
+        # Ruby 2.3.0 is returning a string with ASCII-8BIT encoding.
+        allow(subject).to receive(:url){ 'http://example.com/path/to/%E6%97%A5%E6%9C%AC%E8%AA%9E.txt?bar=baz/fubar'.force_encoding('ASCII-8BIT') }
+      end
+
+      it "should extract correct part" do
+        expect(subject.filename).to eq('日本語.txt')
+      end
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


spec/storage/fog_spec.rb
```diff
@@ -36,5 +36,16 @@
         expect(subject.filename).to eq('foo.txt')
       end
     end
+
+    context "when url contains multi-byte characters ('日本語') in query string" do
+      before do
+        # Ruby 2.3.0 is returning a string with ASCII-8BIT encoding.
```
thomasfedb: This comment only confuses me. Why is the forced encoding required, and is this comment needed?

thomasfedb:
See comment above. Otherwise LGTM. :sparkles:

spec/storage/fog_spec.rb
koic: Thank you for your reply.

Ruby 2.2.4 is returning UTF-8 encoding, but Ruby 2.3.0 is returning ASCII-8BIT.

This `force_encoding('ASCII-8BIT')` means a multi-byte characters that encoded in ASCII-8BIT was the problem in `URI.decode`. This test code to reproduced it.

But use of `CGI.unescape` will be no longer this problem.

Would you like to delete this test code If you confusing. How you do you feel about it?

thomasfedb: > Ruby 2.2.4 is returning UTF-8 encoding, but Ruby 2.3.0 is returning ASCII-8BIT.

Returning from what?

koic: It is `URI::Generic#to_s`. Used by the dependent library fog-aws.

https://github.com/fog/fog-aws/blob/v0.8.0/lib/fog/aws/storage.rb#L309

Does that answer your question?

koic:
Thank you for handling the matter :pray:

PR #1829, opened 2016-01-09
Replace ShamRack by WebMock in specs

mehlah:
WebMock provides a cleaner API that helps to simplify some specs
examples.

This PR comes as a follow up to this discussion https://github.com/carrierwaveuploader/carrierwave/pull/1824#discussion_r48693179


thomasfedb:
Obviously some of these specs need further work, but as a like-for-like replacement I'm happy with this.

PR #1830, opened 2016-01-09
Adding header bearer by url params when use remote_url method

gearnode:
With this feature it is possible to download file protected with Bearer (exemple Slack files)

```ruby

m = Model.new
m.remote_<field>_url = "https://www.secure-baerer.com/test.png?access_token=VALID_TOKEN"
m.save
```

Now this make call with header:

```Authorization: Bearer VALID_TOKEN```


thomasfedb:
Hey @gearnode, my understanding is that generally bearer tokens are passed in the URI, rather than as a header. Can you provide a concrete example of the use-case for this functionality?

Additionally, it doesn't seem to me that the presence of `access_token` as a URI parameter necessarily dictates that the request should be sent with an `Authorization` header. I could use that parameter name in any application I wanted tomorrow.

Finally, is this logic generic and widespread enough to merit inclusion in CarrierWave itself? Let me know your thoughts.

gearnode:
Hey @thomasfedb, you are right. But many API doesn't respect Bearer RFC, for exemple the recent update of slack api about the file gestion, or Zendesk. This API require ```Authorization``` header and do not take into account URI params ```access_token```. 

Bearer is mostly widespread with the header (for the security reason).

Finally, I think when used the URI params ```access_token``` and adding this in header is like a convention.

(Sorry for my poor english)

thomasfedb:
@gearnode could you please direct me to where the examples you note are documented?

Also I disagree that there is a universal convention to translate any URI parameter into a HTTP header. Do any browsers provide this behaviour?

gearnode:
@thomasfedb Edition to my previous post: the pull request is only for the use case on slack. They don't allow external developers to access images with the ```access_token``` in the URI. A developer inside Slack told us they do this for security reasons.

You can find some documentation here: 
- https://medium.com/slack-developer-blog/important-changes-to-files-in-the-web-api-eb38f2a9c1e7#.3silffo3j
- https://api.slack.com/types/file

Do not hesitate to question me again if you find my answer incomplete.

thomasfedb:
@gearnode Thanks for clarifying. I don't think it's appropriate to hardwire CarrierWave to handle these Slack URLs. You should handle this directly in your application or in your wrapper for the Slack API.

PR #1833, opened 2016-01-11
Replace instance variables by `let` in `sanitized_file_spec.rb`

mehlah:
@thomasfedb, as discussed in #1824, here is a sample of what would look like getting the rid of ivars from specs in favor of `let`.
Is this on the right track? Do you want me to do the for the other specs?

### DIFF
--- a/spec/sanitized_file_spec.rb
+++ b/spec/sanitized_file_spec.rb
@@ -8,7 +8,6 @@ rescue LoadError
 end
 
 describe CarrierWave::SanitizedFile do
-
   before do
     FileUtils.cp(file_path('test.jpg'), file_path('llama.jpg'))
   end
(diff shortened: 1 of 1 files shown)


thomasfedb:
Hey @mehlah, I think these are good changes in the right direction.

spec/sanitized_file_spec.rb
```diff
@@ -20,21 +19,23 @@
     FileUtils.rm_rf(public_path)
   end
 
-  describe '#empty?' do
-
+  describe "#empty?" do
     it "should be empty for nil" do
```
thomasfedb: I feel these examples would be better written more like:

```ruby
describe "#empty?" do
  let(:sanitized_file) { CarrierWave::SanitizedFile.new(file) }

  subject {  sanitized_file.empty? }

  context "when SanitizedFile initialized with nil"
    let(:file) { nil }

    specify { expect(subject).to be_true }
  end

  context "when SanitizedFile initialized with an empty string"
    let(:file) { "" }

    specify { expect(subject).to be_true }
  end

  # ...  
end
```

The `specify { expect(subject).to ...` can also be written as `it {  should ...`. I also have a gem (`rspec-expect_it`) that allows writing `specify { expect_it.to ...`.

spec/sanitized_file_spec.rb
```diff
@@ -59,322 +60,337 @@
...
-    before do
-      @sanitized_file = CarrierWave::SanitizedFile.new(nil)
-    end
+  describe "#filename" do
+    let(:sanitized_file) { CarrierWave::SanitizedFile.new(nil) }
 
     it "should default to the original filename if it is valid" do
-      expect(@sanitized_file).to receive(:original_filename).at_least(:once).and_return("llama.jpg")
-      expect(@sanitized_file.filename).to eq("llama.jpg")
+      expect(sanitized_file).to receive(:original_filename).at_least(:once).and_return("llama.jpg")
```
thomasfedb: We seem to rely on mocking a fair bit in this spec. Could you not instead use a `Tempfile`?

```ruby
describe "#filename" do
  let(:sanitized_file) { CarrierWave::SanitizedFile.new(file) }

  subject { sanitized_file.filename }

  context "when the original filename is valid"
    let(:file) { Tempfile(["image", ".jpg"]) }
    let(:filename) { file.path.split("/").last }

    specify { expect(subject).to eq filename }
  end

  context "when the original filename contains illegal characters"
    let(:file) { Tempfile(["a%b#c?d", ".jpg"]) }
    let(:filename) { file.path.split("/").last }

    specify { expect(subject).to eq filename.gsub("a%b#c?d", "a_b_c_d") }
  end
end

thiagofm:
@mehlah @thomasfedb 

I'm currently refactoring specs and I definitely like the approach. I'm also refactoring in other ways, such as removing "should..." from messages, removing duplication... If you are up to it, take a look at my current commits to master and "let's" take the carrierwave specs quality to the next level. Feel free to also suggest improvements over the files I'm refactoring.

Issue #1826, opened 2016-01-05
Code style inconsistencies


mehlah:
CarrierWave code got a lot of style inconsistencies (quotes styles, line breaks, line max lengths, indentations, use of `and` and `or` without a valid reason...)
I think the project quality will benefit from linting the code based on a documented code style guide for contributors.
The ideal scenario is to automate this thanks to Rubocop and even better, to [Hound](https://houndci.com/) (based on Rubocop).
PRs would be automatically reviewed against the shared style guide.
I can take care of documenting this and setting up the rubocop.yml (or hound.yml with our set of rules).
We shouldn't commit a huge cleanup to the code base in one pass to avoid breaking up blames, but from now, pushing (and fixing in passing) a consistent code (I tended to do this in my recent PRs).

Thoughts?

thomasfedb:
I'd be happy to give Hound a shot. My only concern is to avoid putting people off contributing to the project.

thomasfedb:
@bensie what do you think?

mehlah:
> My only concern is to avoid putting people off contributing to the project.

Precisely, I think getting an automated and quick code style review would make contributions easier to accept/decline. The improved code quality and consistency will make it easier for someone unfamiliar with CW code to dig it in the future.
If we find Hound is too invasive, we could disable it and only let core contributors to use Rubocop in their machines with shared rules in rubocop.yml configuration file.

bensie:
:-1: to Hound, but I'm open to a rubocop.yml file to encourage consistency. PRs should not be rejected based on these style issues though.

thomasfedb:
It does look like hound is fairly invasive. Let's go with putting together a rubocop.yml.

plribeiro3000:
:+1: to try hound

Em sex, 8 de jan de 2016 21:58, Thomas Drake-Brockman <
notifications@github.com> escreveu:

PR #1834, opened 2016-01-11
Add a `rubocop.yml` file with basic cops enabled

mehlah:
As suggest at #1826, I started a `rubocop.yml` file where we could agree on some defaults.
I'll update it based on your comments/votes.
I intentionally disabled some cops that I don't think are very crucial to the project at the moment.
We could tweak those progressively based on how sticking to a style guide would work well for the contributors and core team 
:dancer: :heart:

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,262 @@
+Style/CollectionMethods:
+  Description: Preferred collection methods.
+  StyleGuide: https://github.com/bbatsov/ruby-style-guide#map-find-select-reduce-size
+  Enabled: true
+  PreferredMethods:
+    collect: map
+    collect!: map!
+    find: detect
+    find_all: select
+    reduce: inject
+
+Style/DotPosition:
+  Description: Checks the position of the dot in multi-line method calls.
+  StyleGuide: https://github.com/bbatsov/ruby-style-guide#consistent-multi-line-chains
+  Enabled: true
+  EnforcedStyle: trailing
+  SupportedStyles:
+  - leading
+  - trailing
+
+Style/PredicateName:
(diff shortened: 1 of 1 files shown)


.rubocop.yml
```diff
@@ -0,0 +1,262 @@
...
+  Description: Avoid single-line methods.
+  StyleGuide: https://github.com/bbatsov/ruby-style-guide#no-single-line-methods
+  Enabled: true
+  AllowIfMethodIsEmpty: true
+
+Style/StringLiterals:
+  Description: Checks if uses of quotes match the configured preference.
+  StyleGuide: https://github.com/bbatsov/ruby-style-guide#consistent-string-literals
+  Enabled: true
+  EnforcedStyle: double_quotes
```
mehlah: I preferred double quotes as the default as after looking at the current code base, it seems we have more double quoted strings than single quoted.

.rubocop.yml
```diff
@@ -0,0 +1,262 @@
...
+
+Style/IfUnlessModifier:
+  Description: Favor modifier if/unless usage when you have a single-line body.
+  StyleGuide: https://github.com/bbatsov/ruby-style-guide#if-as-a-modifier
+  Enabled: false
+  MaxLineLength: 80
+
+Style/OptionHash:
+  Description: Don't use option hashes when you can use keyword arguments.
+  Enabled: false
```
mehlah: I disabled this as we want Ruby 2.0 support, but required keyword args (when no default values make sense) are available starting from Ruby 2.1

thomasfedb:
Looks like a good set of settings to start with. What sort of output does this produce when run against the current codebase?

PR #1847, opened 2016-01-27
Fix 0.10 stable build

### LINKED ISSUE #1822: Fix 0.10-stable build
As the master branch is now targetting 1.0.0, and a large number of applications use CarrierWave 0.10.0, we should ensure that the `0.10-stable` branch builds successfully so that we have a basis for any future fixes required in the 0.10.x series.

mehlah:
This PR should fix the build for 0.10-stable branch.
It's a first step I hope towards releasing an intermediate version with some backports, before the 1.0 release ✌️.
@thomasfedb I'd be happy to discuss and share some thoughts on the release plan you guys wants to go with.

Fix #1822

### DIFF
--- a/carrierwave.gemspec
+++ b/carrierwave.gemspec
@@ -36,6 +36,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "sham_rack"
   s.add_development_dependency "fog", "~> 1.20.0"
   s.add_development_dependency "unf"
+  s.add_development_dependency "net-ssh", "~> 2.9.0"
   s.add_development_dependency "mini_magick", ">= 3.6.0"
   s.add_development_dependency "rmagick"
   s.add_development_dependency "nokogiri", "~> 1.5.10" # 1.6 requires ruby > 1.8.7
--- a/carrierwave.gemspec
+++ b/carrierwave.gemspec
@@ -35,6 +35,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "rspec", "~> 2.13.0"
   s.add_development_dependency "sham_rack"
   s.add_development_dependency "fog", "~> 1.20.0"
+  s.add_development_dependency "unf"
   s.add_development_dependency "mini_magick", ">= 3.6.0"
   s.add_development_dependency "rmagick"
   s.add_development_dependency "nokogiri", "~> 1.5.10" # 1.6 requires ruby > 1.8.7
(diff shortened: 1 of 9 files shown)


thomasfedb:
Appears to be some issue installing `i18n` on CI.

mehlah:
@thomasfedb the build fails for older Rubies (1.8 and 1.9.2) due to the gems dependencie https://travis-ci.org/carrierwaveuploader/carrierwave/jobs/105075508#L189
I guess we could fix that with something like `gem 'i18n', '~> 0.6.0', :platforms => [:mri_18, :mri_19]` but I'm not sure if we want to support Ruby < 1.9.3.
Let me know what do you think.

thomasfedb:
Given that Rails 3.2 supports Ruby 1.8.7, and also that we're fixing this retrospectively, I feel that we should continue to test 0.10.x against 1.8.7, although we could drop support in 0.11.x if that is released.

thomasfedb:
Given that `activesupport` should have a dependency on an appropriate version of `i18n` I'm curious how this issue arises...

carrierwave.gemspec
```diff
@@ -29,12 +29,14 @@ Gem::Specification.new do |s|
...
-  s.add_development_dependency "mysql2"
+  s.add_development_dependency "pg"
   s.add_development_dependency "rails", ">= 3.2.0"
   s.add_development_dependency "cucumber", "~> 1.3.2"
   s.add_development_dependency "rspec", "~> 2.13.0"
   s.add_development_dependency "sham_rack"
-  s.add_development_dependency "fog", ">= 1.3.1"
+  s.add_development_dependency "fog", "~> 1.20.0"
+  s.add_development_dependency "unf"
+  s.add_development_dependency "net-ssh", "~> 2.9.0"
```
mtsmfm: To fix test for ruby 1.8.7, we should specify some gem versions:

```
s.add_development_dependency "mini_magick", "< 4"
s.add_development_dependency "i18n", "< 0.7"
s.add_development_dependency "mime-types", "< 2"
s.add_development_dependency "sprockets", "< 3"
s.add_development_dependency "net-ssh", "< 2.10"
s.add_development_dependency "rack-cache", "< 1.3"
```

spec/spec_helper.rb
```diff
@@ -23,8 +21,16 @@
   Fog.mock!
 end
 
+module Rails
```
mtsmfm: To `require 'rails'`, we don't need this workaround :smile:

mehlah: Thanks for the hint. I was on the wrong track ;-)

mehlah:
Annd it's :green_heart: ! Thanks @mtsmfm for your hints.

mtsmfm:
:tada:

thomasfedb:
Looks great!

PR #1856, opened 2016-01-31
Add docs to README.md about the workaround for #1855

### LINKED ISSUE #1855: Uploading multiple images with same image name doesn't work
## Ideal behavior

Maybe an automatical image name update something `omg.png, omg.png` to `omg.png, omg_2.png`?
## Current behavior

DB column value ends up like `["omg.png", "omg.png"]`, but only the first image is actually uploaded.
Same thing happens when I use the [unique filename strategy](https://github.com/carrierwaveuploader/carrierwave/wiki/How-to%3A-Create-random-and-unique-filenames-for-all-versioned-files#unique-filenames)(same filename but only 1 image uploaded).

BTW I found out that this happens when using iPhone to upload images :)

multiple file upload feature was added in https://github.com/carrierwaveuploader/carrierwave/pull/1481

take:
closes #1855 

Maybe we can use something different than `cache_id`, but confirmed that this will do the job

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,18 @@ u.avatars[0].current_path # => 'path/to/file.png'
 u.avatars[0].identifier # => 'file.png'
 ```
 
+### Workaround for multi file with same name
+
+Some client sends different images with the same file name(confirmed in iPhone6), and the default `store_dir` and `filename` value will not work in this case.
+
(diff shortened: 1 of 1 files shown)


thomasfedb:
I'd rather fix the issue rather than providing a workaround. Perhaps add this as a comment in #1855.

PR #1860, opened 2016-02-01
Fix returning a new instance of the uploader on every uploader call

DarthSim:
How to reproduce:
```ruby
class Picture < ActiveRecord::Base
  mount_uploader :image, ImageUploader
end

pic = Picture.new
pic.image.object_id # => 70317811568320
pic.image.object_id # => 70317821530860
```

First of all, this bug generates new objects every time, and this is definitely not good. Also this affects on actions with uploader. For example:
```ruby
pic = Picture.new
pic.image.retrieve_from_cache!("1454349370-83816-0001-5310/image.jpg")
pic.image.blank? # => true
```

### DIFF
--- a/lib/carrierwave/mount.rb
+++ b/lib/carrierwave/mount.rb
@@ -139,7 +139,7 @@ module CarrierWave
       mod.class_eval <<-RUBY, __FILE__, __LINE__+1
 
         def #{column}
-          _mounter(:#{column}).uploaders[0] or _mounter(:#{column}).blank_uploader
+          _mounter(:#{column}).uploaders[0] ||= _mounter(:#{column}).blank_uploader
         end
 
         def #{column}=(new_file)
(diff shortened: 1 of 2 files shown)


mehlah:
LGTM :+1:

thomasfedb:
Looks like a sensible change @DarthSim. My only query before I can merge this is if it keeps the single and plural mounters consistent. Could you check this please?

DarthSim:
@thomasfedb It does, why shouldn't it?

thomasfedb:
@DarthSim I just wanted to compare the behaviour of these two:

```ruby
def mount_uploader(column, uploader=nil, options={}, &block)
  # ...

  def #{column}
    _mounter(:#{column}).uploaders[0] ||= _mounter(:#{column}).blank_uploader
  end
```

```ruby
def mount_uploaders(column, uploader=nil, options={}, &block)
  # ...

  def #{column}
    _mounter(:#{column}).uploaders
  end
```

thomasfedb:
Clearly you changes makes this *more* consistent, as the later example was already returning a stable object.

PR #1863, opened 2016-02-02
Changes to Mounter and MiniMagick

oss92:
Changes in MiniMagick#resize_to_fill and introduced Mounter#init_uploaders that result in less code complexity and less duplication.

### DIFF
--- a/lib/carrierwave/mounter.rb
+++ b/lib/carrierwave/mounter.rb
@@ -38,11 +38,7 @@ module CarrierWave
 
     def cache(new_files)
       return if not new_files or new_files == ""
-      @uploaders = new_files.map do |new_file|
-        uploader = blank_uploader
-        uploader.cache!(new_file)
-        uploader
-      end
+      init_uploaders(new_files, :cache!)
 
       @integrity_error = nil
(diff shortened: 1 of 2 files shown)


thomasfedb:
Hey @oss92. I think we need to be very careful to not confuse *"less code"* with *"less complexity"*.

I don't like your changes to `mounter.rb` as they introduce an extra degree of indirection, and make understanding the result of the method calls more difficult.

Whilst I agree that there's something less than optimal about having three similar methods, I'm convinced that this isn't the solution. I haven't been able to come up with a better solution than the existing code yet.

Likewise, with your changes to `mini_magick.rb`, we'd be swapping 11 lines of fairly sparse, somewhat repetitive, dull code for 4 lines of highly compact, dense code, which also makes use of a number of more advanced constructs.

In either case, I'm not able to justify the change. Thank you for your suggestions, they are much appreciated. It would be nice to resolve both these trouble-spots in our codebase, but I think we need to look further for elegant, simple, solutions.

I'm going to close this PR, but feel free to open an issue to discuss the problem at hand.

PR #1865, opened 2016-02-03
ActiveRecord test for nested attributes

oss92:
Increased AR test coverage by adding a test for setting `remove_#{column}` using nested_attributes. (Uncovered area)

It did not seem to work on 0.10 (http://stackoverflow.com/questions/35165619/remove-carrierwave-file-from-associated-model). I tried it using nested attributes and if only the `remove_#{column}` is set to true (` {remove_image => true}`), it did not update the record. If it was ` {remove_image => true, another_attribute => "CHANGE"}` it worked. 

It seems to be fixed on the master branch so I wrote a test to verify that and prevent any future mixups.

### DIFF
--- a/spec/orm/activerecord_spec.rb
+++ b/spec/orm/activerecord_spec.rb
@@ -15,6 +15,10 @@ def drop_table(name)
   ActiveRecord::Base.connection.drop_table(name)
 end
 
+def add_has_many_relation(parent_table, children_table)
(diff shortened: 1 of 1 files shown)


thomasfedb:
Hi @oss92, I don't think this is a CarrierWave issue. Rails is responsible for assigning attributes to the correct model instances when handling nested attributes.

Given that there are no issues with this behaviour on the master branch, it's unknown if adding these tests would prevent a regression in any case.

PR #1866, opened 2016-02-03
Backport #1797 to 0.11-stable

mtsmfm:
Backport #1797 to 0.11-stable

------------------

Add counter to cache_id

Old version of generate_cache_id was collision-prone: for complex factories with multiple uploaders we had quite a few flaky specs because of that. With a counter, this (and similar situations like https://github.com/carrierwaveuploader/carrierwave/issues/1232) should be resolved completely.

### DIFF
--- a/features/caching.feature
+++ b/features/caching.feature
@@ -23,6 +23,6 @@ Feature: uploader with file storage
   Scenario: retrieving a file from cache
     Given an uploader class that uses the 'file' storage
     And an instance of that class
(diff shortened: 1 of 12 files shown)


thomasfedb:
Thanks @mtsmfm!

mtsmfm:
:wink:

mtsmfm:
@thomasfedb When will 0.11 be released?
I think this fix is very important for stable gem users.

thomasfedb:
We haven't committed to any definite timeframe at this point @mtsmfm. Sorry I can't be more specific.

Issue #1602, opened 2015-03-18
Version from original file instead of other version


tomek-railwaymen:
I have the following code in my uploader:

``` ruby
  version :large do
    process :convert_to_png

    def full_filename(for_file = model.pdf.file)
      "large-#{File.basename(for_file, '.pdf')}.png"
    end

    def convert_to_png
      manipulate! do |img|
        img.format 'png' do |cmd|
          cmd.merge! ['-density', '200', '-quality', '100']
        end
        img = yield(img) if block_given?
        img
      end
    end
  end

  version :thumb, from_version: :large do
    process resize_to_fit: [200, 200]

    def full_filename(for_file = model.pdf.file)
      "thumb-#{File.basename(for_file, '.pdf')}.png"
    end
  end
```

Unfortunately `thumb` version is created based on original file (always in PDF format). That file is broken because it is PDF as well (instead of PNG, like `large` version). 

I noticed that `large` version is generated by command like `convert /tmp/mini_magick20150318-4571-1ma89u5.pdf[0] -density 200 -quality 100 /tmp/mini_magick20150318-4571-ifujor.png` and `thumb` uses PDF file: `mogrify -resize 200x200 /tmp/mini_magick20150318-4571-5or0in.pdf`

cipater:
+1 `from_version` doesn't appear to be working. Which is to say that all versions look to be built off the original rather than version specified by `from_version`.

Given
```ruby
  process resize_to_limit: [1600, 1600]
  process :store_geometry

  version :optimized do
    process :pngquant
    process :optimize
    process :store_geometry
  end

  version :large, from_version: :optimized do
    process resize_to_limit: [1400, 1400]
    process :store_geometry
  end
```

The original file is 2.5MB. The optimized version is 900KB. And the large version is 1.8MB.

If the large version were in fact built from the optimized version, I would expect it to be smaller than 900KB, not twice the size.

cianoswift:
Any updates on this? I'm still having this issue on v1.1.0

mshibuya:
Fixed by #1879.

PR #1879, opened 2016-02-16
regard :from_version option on calling #recreate_versions! with args

### LINKED ISSUE #1164: `recreate_versions!(...)` ignores `:from_version => ...`
Having the following code:

``` ruby
  version :medium do
    ...
  end

  version :small, :from_version => :medium do
    ...
  end
end
```

CarrierWave works as expected when I assign the file for the first time.
It also works when I call `recreate_versions!` without any parameters.
But when I call `recreate_versions!(:small)`, it ignores `:from_version` and use the original file that was uploaded.

I suspect the bug to be in the method `recreate_versions!`:

``` ruby
def recreate_versions!(*versions)
  # Some files could possibly not be stored on the local disk. This
  # doesn't play nicely with processing. Make sure that we're only
  # processing a cached file
  #
  # The call to store! will trigger the necessary callbacks to both
  # process this version and all sub-versions
  if versions.any?
    file = sanitized_file if !cached?
    store_versions!(file, versions)
  else
    cache! if !cached?
    store!
  end
end
```

https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/uploader/versions.rb#L223

It passes `sanitized_file` to `store_versions!` which I suspect is the original file that was uploaded instead of being the cache version corresponding to `:medium`

hedgesky:
Fixes #1164. Maybe it's also related to #1602.

### DIFF
--- a/README.md
+++ b/README.md
@@ -518,26 +518,6 @@ User.find_each do |user|
 end
 ```
 
-### Recreating versions with `:from_version` dependency
-
-Imagine you have a version with `:from_version`:
-
-```ruby
-class MyUploader < CarrierWave::Uploader::Base
-
-  version :thumb do
-    process resize_to_fill: [280, 280]
-  end
-
-  version :small_thumb, from_version: :thumb do
-    process resize_to_fill: [20, 20]
-  end
-
-end
-```
-
-In this case, when you call `recreate_versions!(:small_thumb)`, both `:small_thumb` and `:thumb` versions will be recreated.
-
 ## Configuring CarrierWave
 
 CarrierWave has a broad range of configuration options, which you can configure,
--- a/README.md
+++ b/README.md
@@ -518,6 +518,26 @@ User.find_each do |user|
 end
 ```
 
+### Recreating versions with `:from_version` dependency
+
+Imagine you have a version with `:from_version`:
+
+```ruby
+class MyUploader < CarrierWave::Uploader::Base
+
+  version :thumb do
+    process resize_to_fill: [280, 280]
+  end
+
+  version :small_thumb, from_version: :thumb do
+    process resize_to_fill: [20, 20]
+  end
+
+end
+```
+
+In this case, when you call `recreate_versions!(:small_thumb)`, both `:small_thumb` and `:thumb` versions will be recreated.
+
 ## Configuring CarrierWave
 
 CarrierWave has a broad range of configuration options, which you can configure,
--- a/lib/carrierwave/uploader/versions.rb
+++ b/lib/carrierwave/uploader/versions.rb
@@ -223,14 +223,10 @@ module CarrierWave
 
         if names.any?
           file = sanitized_file
-          # We should somehow filter versions to cache in #cache_versions!.
-          # Passing arguments through all of those callbacks is difficult, so @versions_to_cache
-          # instance variable is introduced. It's used only here and in #cache_versions!.
-          @versions_to_cache = substitute_dependent_with_parents(names)
-          versions_to_store = (@versions_to_cache + names).uniq
+          set_versions_to_cache_and_store(names)
           cache!(file)
-          @versions_to_cache = nil
-          store_versions!(file, versions_to_store)
+          store!(file)
+          reset_versions_to_cache_and_store
         else
           cache! if !cached?
           store!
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -518,6 +518,26 @@ User.find_each do |user|
 end
 ```
 
+### Recreating versions with `:from_version` dependency
```
thiagofm: Please move this to the wiki, the Readme is already too big for a very specific user case

lib/carrierwave/uploader/versions.rb
```diff
@@ -230,6 +238,24 @@ def recreate_versions!(*versions)
       end
 
     private
+
+      # To correctly cache dependent version we need to be sure that its parent is cached, too.
+      # But caching parent version automatically cache its descendents.
+      # So, while recreating certain versions in #cache_versions! we should skip dependent versions
+      # as soon as their caching will be automatically started by parent.
+      def substitute_dependent_with_parents(names)
+        with_dependencies = []
```
thiagofm: You can convert this to a inject

lib/carrierwave/uploader/versions.rb
```diff
@@ -213,16 +213,24 @@ def url(*args)
...
-        if versions.any?
-          file = sanitized_file if !cached?
-          store_versions!(file, versions)
+
+        if names.any?
+          file = sanitized_file
+          # We should somehow filter versions to cache in #cache_versions!.
+          # Passing arguments through all of those callbacks is difficult, so @versions_to_cache
+          # instance variable is introduced. It's used only here and in #cache_versions!.
+          @versions_to_cache = substitute_dependent_with_parents(names)
```
thiagofm: I would avoid code like this at the current moment. I would find a way to extract this or improve the current method. I know that there might be others places where we are instantiating variables outside of the constructor of the class, but as of now, I see this as a bad smell and should be avoided at all costs.

thiagofm:
@hedgesky I've added a couple of comments, I would appreciate very much if you go forward with it, there's definitely some value laying there. :+1:

lib/carrierwave/uploader/versions.rb
```diff
@@ -213,23 +213,60 @@ def url(*args)
...
         else
           cache! if !cached?
           store!
         end
       end
 
     private
+
+      def set_versions_to_cache_and_store(names)
+        @versions_to_cache = source_versions_of(names)
```
hedgesky: Basically, it's still an instance variable. And now even two of them instead of one 😄 . But introducing them helps to avoid conditions in `cache_versions!` and `store_versions!` methods. Also, assigning these variables is hidden in methods with readable names. So, I think this change decreases overall complexity.

lib/carrierwave/uploader/versions.rb
```diff
@@ -213,23 +213,60 @@ def url(*args)
...
+        @versions_to_store = active_versions_with_names_in(@versions_to_cache + names)
```
hedgesky: Basically, it's still an instance variable. And now even two of them instead of one 😄 . But introducing them helps to avoid conditions in `cache_versions!` and `store_versions!` methods. Also, assigning these variables is hidden in methods with readable names. So, I think this change decreases overall complexity.

lib/carrierwave/uploader/versions.rb
```diff
@@ -263,19 +300,14 @@ def full_original_filename
...
 
       def cache_versions!(new_file)
-        dependent_versions.each do |name, v|
+        versions_to_cache.each do |name, v|
           v.send(:cache_id=, @cache_id)
           v.cache!(new_file)
         end
       end
 
-      def store_versions!(new_file, versions=nil)
```
hedgesky: Now versions to store is part of object's state, so we don't need extra argument here.

lib/carrierwave/uploader/versions.rb
```diff
@@ -213,23 +213,60 @@ def url(*args)
...
         # process this version and all sub-versions
-        if versions.any?
-          file = sanitized_file if !cached?
-          store_versions!(file, versions)
+
+        if names.any?
+          file = sanitized_file
+          set_versions_to_cache_and_store(names)
+          cache!(file)
+          store!(file)
```
hedgesky: Now we avoid duplication and possible errors by using similar approach to cache/store in both branches of a condition.

hedgesky:
@thiagofm Thanks for looking into that! Following your suggestions, I've made some changes. As for me, it's much readable now.

lib/carrierwave/uploader/versions.rb
thiagofm: Those instance variables being set here makes me think that the recreate_versions could be in its own class. As of now, I guess it's good to avoid adding more code here. Do you mind giving it a try @hedgesky ?

hedgesky:
@thiagofm Sorry for late answer. I'll try it when I have spare time. I agree with you that this functionality should be extracted but don't see right now how it should be implemented properly. Will think about it.

mshibuya:
Merging in finally, thanks!

antonakospanos:
When are we going to have this released?

lib/carrierwave/uploader/versions.rb
rogerhu: We noticed this change causes an AWS S3-backed store to trigger callbacks and send DELETES.  Is this expected behavior?

rogerhu: Seems related to https://github.com/carrierwaveuploader/carrierwave/issues/2385

rogerhu: Found the reason -- this change caused the cache storage to match the uploader:

https://github.com/carrierwaveuploader/carrierwave/commit/629afecbaeccd2300e4660b78ee36bd95dd845c5

PR #1882, opened 2016-02-21
Remove json gem

deepj:
JSON support is in Ruby since 1.9 so there is no need to have json gem as a dependency.

Inspired by this Rack’s PR https://github.com/rack/rack/pull/1011


thomasfedb:
Thanks @deepj, Looks good, I'll have a closer look a little later, any very likely get this merged today.

deepj:
@thomasfedb Thanks!

PR #1895, opened 2016-03-11
improve the performance to access AWS and Google

kevinclcn:
When using :AWS or :Google in the configuration file, access the images stored on AWS or Google are very slow. This is because when using :AWS or :Google, this public_url will return nil, and carrierwave will retrive the public_url by a service call.


kevinclcn:
The test failure has nothing to do with my change.

thiagofm:
@kevinclcn I have rerun the specs, please do it next time. All is green!

kevinclcn:
@thiagofm thank you, I just get how to rerun the specs.

PR #1561, opened 2015-02-13
Don't error when size is called on a deleted file

danielevans:
When using Fog and AWS and a new file is uploaded and the old file is missing  `content_length` is called on nil, resulting in the inability to upload a new file.

This change updates the `CarrierWave::Storage::Fog` class such that it will not error if this situation occurs and instead return a 0 file size.

It seems that this change could also be made in https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/sanitized_file.rb#L136 by reversing the conditional `(self.size.zero? && ! self.exists?)`, however the change in this PR seemed less invasive.

The error caused by this issue:
````
NoMethodError: undefined method `content_length' for nil:NilClass
	from /srv/app/releases/20150213002036/vendor/bundle/ruby/2.1.0/bundler/gems/carrierwave-0033279be9b6/lib/carrierwave/storage/fog.rb:291:in `size'
	from /srv/app/releases/20150213002036/vendor/bundle/ruby/2.1.0/bundler/gems/carrierwave-0033279be9b6/lib/carrierwave/uploader/proxy.rb:57:in `size'
	from /srv/app/releases/20150213002036/vendor/bundle/ruby/2.1.0/bundler/gems/carrierwave-0033279be9b6/lib/carrierwave/sanitized_file.rb:96:in `size'
	from /srv/app/releases/20150213002036/vendor/bundle/ruby/2.1.0/bundler/gems/carrierwave-0033279be9b6/lib/carrierwave/sanitized_file.rb:136:in `empty?'
	from /srv/app/releases/20150213002036/vendor/bundle/ruby/2.1.0/bundler/gems/carrierwave-0033279be9b6/lib/carrierwave/uploader/cache.rb:111:in `cache!'
	from /srv/app/releases/20150213002036/vendor/bundle/ruby/2.1.0/bundler/gems/carrierwave-0033279be9b6/lib/carrierwave/mounter.rb:43:in `block in cache'
	from /srv/app/releases/20150213002036/vendor/bundle/ruby/2.1.0/bundler/gems/carrierwave-0033279be9b6/lib/carrierwave/mounter.rb:41:in `map'
	from /srv/app/releases/20150213002036/vendor/bundle/ruby/2.1.0/bundler/gems/carrierwave-0033279be9b6/lib/carrierwave/mounter.rb:41:in `cache'
[...]

PR #1915, opened 2016-04-07
Don't raise an error when content_type is called on a deleted file

jvenezia:
Related to https://github.com/carrierwaveuploader/carrierwave/pull/1561
Same problem, with `content_type`.


mshibuya:
Thanks!

PR #1921, opened 2016-04-14
Update mini_magick.rb to use with jruby

thanqminh:
default value declaration for the parameter  `combine_options` should be defined as `combine_options={}` to be valid in jruby instead of `combine_options: {}`


thiagofm:
Hello @thanqminh, that change breaks both the API of carrierwave and also a lot of tests.

thiagofm:
The issue at hand seems to be: https://github.com/jruby/jruby/issues/3760

PR #1925, opened 2016-04-21
replaces requiring all of fog with requiring just the supported providers

langalex:
this way, an app can depend on only for example fog-aws, instead of all 24 fog provider gems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ notifications:
   email: false
 
 rvm:
- - 1.9.3
  - 2.0
  - 2.1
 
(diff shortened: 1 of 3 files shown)


thiagofm:
@langalex I like this change. But we're currently having issues with the mime-types-data gem in that branch.

I know that it's very unlikely that this would break anything, but I would like to have this problem fixed before we merge further things there.

If possible, it would be good if you could find a solution for the mime-types-data issue that breaks our build.

langalex:
ok no rush.

your mime-types problem seems to be this: `Gem::InstallError: mime-types-data requires Ruby version >= 2.0`. (output during bundler). ditch pre 2.0 rubies?

thiagofm:
@langalex mind removing 1.9.3 from .travis.yml of your PR to see if everything is green? As it 0.11 doesn't seem to be working at all for 1.9.3, might be worth to remove the compatibility there.

langalex:
@thiagofm done and all is green.

thiagofm:
@langalex thanks a lot for taking your time to fix this issue!

Issue #1933, opened 2016-05-04
ImageMagick exploit: cve-2016-3714 "magic byte" validation


timcheadle:
[This article](https://imagetragick.com/) alerts that a new ImageMagick exploit is being used in the wild and describes the following fixes:
- Using a policy file to limit possible attack vectors
- Checking the "magic bytes" at the beginning image files to validate that they are actually image files

Can someone provide an update as to whether or not this affects CarrierWave and, if so, what steps are being taken or have been taken to mitigate the vulnerability?

d4rky-pl:
Carrierwave uses [mime-types](https://github.com/mime-types/ruby-mime-types) gem which detect content type based on filename. I've already flagged this as potential security issue over a year ago (#1543) but the consensus at the time was that it's outside the scope of this project and a custom processor would be preferable.

thiagofm:
@d4rky-pl given the current issue, do you mind giving it a try with your fix? I was also looking at a solution yesterday, but didn't know about the file from unix solution.

I would like to leave this outside carrierwave, but given the current issue...

We'll do what we can to protect our users from exploits. If it has to be some sort of temporary patch, I can take care of maintaining it, no problem.

fqueiruga:
What approaches are being considered to detect the content type? I'd use something like [mimemagick](https://github.com/minad/mimemagic) rather than calling the file command. The reason for this is that in some production environments these kind of syscalls are disabled.

What are your thoughts on this?

namxam:
From what I read, this is also the approach paperclip is using: https://github.com/thoughtbot/paperclip/blob/v4.3.6/lib/paperclip/content_type_detector.rb#L69-L72

thiagofm:
The mimemagick approach seems good, anybody is free to submit a PR with a solution. 👍

alexmuller:
We've confirmed that a Rails application using carrierwave 0.9.0 is vulnerable to CVE-2016-3714 with ImageMagick `8:6.7.7.10-6ubuntu3`.

locriani:
I created a PR to fix this issue: https://github.com/carrierwaveuploader/carrierwave/pull/1934

fqueiruga:
Will there be a backport for older versions of this fix?

marksweston:
Anyone looking for immediate solutions might well take a look at https://github.com/DarthSim/carrierwave-bombshelter

thiagofm:
@fqueiruga yes there will be a backport for 0.11

[...]

PR #1934, opened 2016-05-04
Mitigate against CVE-2016-3714 (ImageTragick)

locriani:
- Use the MimeMagic gem to detect content types via file headers
- Fallback to Mime::Types if the file does not exist
- Make the ruby.gif file actually a GIF, so blacklist tests pass (was previously a png masquerading as a gif)
- Update the README to detail all required steps to mitigate against CVE-2016-3714


locriani:
Ref: https://github.com/carrierwaveuploader/carrierwave/issues/1933

locriani:
Issues:
- This only allows the end user to mitigate by specifying a white list of allowable content types. Should this be addressed in the readme, or by adding a default white list?
- I do not know the code path that's resolved at [line 268](https://github.com/carrierwaveuploader/carrierwave/pull/1934/files#diff-6599f0c1ad429a2d9d0f8cdc32926e8fL268), this could be a vector for an inaccurate mime type

namxam:
Regarding issue 1, I would vote for a default whitelist. Having a secure setup should always be default. But there should be some info in the README how to change it and why those defaults have been chosen.

thiagofm:
@locriani can you please remove the images changes on your commit?

locriani:
@thiagofm I can't remove the change to the `ruby.gif` image without breaking these specs:
- spec/uploader/content_type_blacklist_spec.rb:33
- spec/uploader/content_type_whitelist_spec.rb:27

With the changes to the mime type validation, the `ruby.gif` image no longer parses with a mime type of `image/gif` (which is spoofed based on the extension), but as an `image/png` (which is what it actually was). The image change I made converted it into a gif, so the new file header based mime-type detection would still continue to read it as an `image/gif`.

Would you prefer those tests be changed?

thiagofm:
@locriani yeah, perhaps might be useful adapt the tests.

locriani:
@thiagofm Done. I created a new PR for the backport, as well.

thiagofm:
As this is the master branch, I guess it's worth to have a "lasting" solution for this problem. As it's very common for people to have outdated versions of imagemagick(given that there's a fix for the CVE), I think we'll always have to handle this.

@locriani how this approach contrasts with https://github.com/DarthSim/carrierwave-bombshelter ? (uses fastimage gem)

I've merged the backport and will generate a new release for carrierwave 0.11(this is what most of the people are probably/should be using).

locriani:
@thiagofm That's reasonable.

I think it would make sense to make a default whitelist of acceptable mime types, which could include the common image types (but would need to exclude SVG, by default, because of the MSVG issue). This doesn't look like it would be too complex to accomplish with the carrierwave code, and I think it could be a good approach.

The problems I see here from a security perspective are:
1. denial of service from overly-large images
2. script execution via extension names
3. improper / unexpected processing of files as images (this CVE)

I'm not sure 1 would be in scope for carrierwave to handle, as it's most effectively mitigated at the webserver (apache / nginx) level. 2 is already possible to mitigate with carrierwave, but there are no defaults, and 3 is not possible to mitigate until there's inspection of the file headers built into carrierwave.

Bombshelter's approach is to do validation of the content-type of the file in a validation before it even hits CarrierWave. The gem used for header validation there may be better and easier to use than `mimemagic`, though.

thiagofm:
@locriani I was also thinking about something in that direction(default whitelist of mimetypes).

I think the item 1 is definitely out of context.

Might be worth doing a comparison between the two gems(if possible), or perhaps pushing that protection to the RMagick/Minimagick side.

locriani:
Makes sense to me. I'll spend some time comparing the two gems later today when I have some free time.

thiagofm:
Thanks a lot for putting your time to it!

locriani:
It looks like FastImage is actually very limited and unable to detect multiple filetypes that RMagick / ImageMagick supports.

We can't push it over to RMagick / ImageMagick for identification with `identify`, unfortunately, because the identify command has the same RCE flaws.

rgueldem:
It seems like there is still an issue if mimemagic doesn't know the mimetype. In that case it falls back to the possibly spoofed content type...

locriani:
@rgueldem After reviewing the POC and the code again, you're absolutely right. I'll create (yet another) fix for 0.11.0 and revisit this code here.

README.md
```diff
@@ -245,6 +245,22 @@ class NoJsonUploader < CarrierWave::Uploader::Base
...
+
+```ruby
+class MyUploader < CarrierWave::Uploader::Base
+  def content_type_whitelist
+    [/image\//]
+  end
+end
+```
+
+**WARNING**: A `content_type_whitelist` is the only form of whitelist or blacklist supported by CarrierWave that can effectively mitigate against CVE-2016-3714. Use of `extension_type_whitelist` will not inspect the file headers, and thus still leaves your application open to the vulnerability.
```
pjg: There is a typo here. `extension_type_whitelist ` should be just `extension_whitelist`.

lib/carrierwave/sanitized_file.rb
```diff
@@ -329,6 +328,22 @@ def sanitize(name)
...
     end
 
+    def existing_content_type
+      if @file.respond_to?(:content_type) && @file.content_type
+        @file.content_type.to_s.chomp
+      end
+    end
+
+    def mime_magic_content_type
+      MimeMagic.by_magic(File.open(path)).try(:type) if path
```
ryancw: Is there a reason we can't use the `file` object instead as an argument for `by_magic`? I am trying to use this fix to support `StringIO` objects I am storing through CarrierWave, but it does not work due to lack of file path.

obduk: This also causes a file handle to be left open. When upgrading from 0.11.0 to 0.11.1, all my upload tests started to fail as I always check all files are closed https://github.com/obduk/cms/blob/master/spec/support/all/bad_tests.rb#L2

I tracked it down to this code, which as also been backported in to 0.11.1 in #1936 

I don't mind fixing it, but don't want to cause collisions with lines that people are already working on?

thiagofm: Feel free to do it. We still don't have a proper solution. Would appreciate it very much.

FergusonSean:
Was this ever fixed in 1.0? I'm able to get around my content type whitelist by changing file extensions.

lukeasrodgers:
@FergusonSean it is not, AFAICT. Looks like the last commit in common between mainline (current release 1.2.2) and 0.11.2 is 3a0978a. Only the v0.11.2 line has the imagetragick fix.

birthdaycorp:
Hi all! Quick question. The doc says:

> and you MUST either disable ImageMagick's default SVG delegate or use the RSVG delegate for SVG processing.

How does one go about doing this? The first part of the solution is just adding the whitelist, but this part sort of eludes me.

PR #1942, opened 2016-05-07
Do not allow file type spoofing if the header bytes are not recognized

locriani:
This fixes an issue raised in https://github.com/carrierwaveuploader/carrierwave/pull/1934, where a malicious user could bypass the stricter mime-type checking by choosing header bytes that are not recognized by the `mimemagic` gem.

### DIFF
--- a/lib/carrierwave/sanitized_file.rb
+++ b/lib/carrierwave/sanitized_file.rb
@@ -315,7 +315,7 @@ module CarrierWave
     end
 
     def mime_magic_content_type
-      MimeMagic.by_magic(File.open(path)).try(:type) if path
+      MimeMagic.by_magic(File.open(path)).try(:type) || 'invalid/invalid' if path
     rescue Errno::ENOENT
       nil
     end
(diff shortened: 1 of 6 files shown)


myers:
I think you are on the right path here.  One concern I have after testing 0.11.3 in my own app, is that [`existing_content_type`](https://github.com/locriani/carrierwave/blob/072e3c2ad9b67807389fd612ce14c950282af2c9/lib/carrierwave/sanitized_file.rb#L311-L315) trust what `@file` says about itself.  When you are uploading a file this is an instance of [ActionDispatch::Http::UploadedFile](http://api.rubyonrails.org/classes/ActionDispatch/Http/UploadedFile.html) which has the content type set to a value form the browser, which shouldn't be trusted.  I think [`content_type`](https://github.com/locriani/carrierwave/blob/072e3c2ad9b67807389fd612ce14c950282af2c9/lib/carrierwave/sanitized_file.rb#L247-L252) should only trust `mime_magic_content_type`.

PR #1946, opened 2016-05-09
Include the asset_host in default_url if it returns a relative url

mcls:
If #default_url retuns a relative url, then it makes sense to add the
asset_host to it, if configured.

### DIFF
--- a/lib/carrierwave/uploader/default_url.rb
+++ b/lib/carrierwave/uploader/default_url.rb
@@ -3,7 +3,7 @@ module CarrierWave
     module DefaultUrl
 
       def url(*args)
-        super || default_url(*args)
+        super || determine_default_url(*args)
       end
 
       ##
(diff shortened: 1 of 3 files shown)


mshibuya:
I don't feel usefulness of this change so much, just returning absolute url in `default_url` isn't enough?

mcls:
Yes, you're right. Currently we're using something like what's suggested in the default uploader template:

```
def default_url
  ActionController::Base.helpers.asset_path("fallback/" + [version_name, "default.png"].compact.join('_'))
end
```
https://github.com/carrierwaveuploader/carrierwave/blob/a2c93fe1a9d3a3f60d1662e5ec34fdba7bf2b304/lib/generators/templates/uploader.rb#L20

It's been a while, but I think I was confused because I didn't expect a method suffixed with `_path` to return an absolute url. However, the [`asset_path`](http://api.rubyonrails.org/classes/ActionView/Helpers/AssetUrlHelper.html#method-i-asset_path) method does include the `asset_host` if it's set.

PR #1949, opened 2016-05-12
Add note about the used method for multiple files

gaetanm:
I wasted time figuring why I wasn't able to do multiple uploads... It was simply because I haven't noticed that it's a different method which is needed.


thiagofm:
Sorry, the readme is already too big and it's also written there mount_upload(s)

PR #1951, opened 2016-05-14
Added a section in the readme to check for presence of file

Samy-Amar:
The programming school I work uses carrierwave, and this is an issue we get regularly with students, who want to check if a photo was uploaded or not to the model through an 'if' statement. 

The answer is surprisingly hard to find online, it's a hard one to guess at first glance, and I believe adding it to the readme would help a lot of people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,15 +121,8 @@ u.avatar.current_path # => 'path/to/file.png'
 u.avatar_identifier # => 'file.png'
 ```
 
-### Checking for presence
-
-
-To check for presence for a file, use `.file` after the column name.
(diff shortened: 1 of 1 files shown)


thiagofm:
I like the idea, but the current Readme is already too big. Can you optimize it for brevity and "searchability"?

Samy-Amar:
@thiagofm: I've made a second draft that fits in two lines, and should be searchable enough. Tell me what you think

thiagofm:
👍  LGTM

aried3r:
Is this different from doing `u.avatar?`?

PR #1955, opened 2016-05-17
Remove 'config.fog_provider' from the 'Using Amazon S3' section.

FarazPatankar:
Removed the `config.fog_provider = 'fog/aws' ` line from the `Using Amazon S3` section because it leads to an error :

`undefined_method 'fog_provider=' for CarrierWave::Uploader::Base:Class (NoMethodError)`

The setup works completely fine without that line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -652,7 +652,6 @@ You can also pass in additional options, as documented fully in lib/carrierwave/
 
 ```ruby
 CarrierWave.configure do |config|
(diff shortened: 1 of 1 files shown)


LIYINGZHEN:
Do you try using master branch ?

`gem 'carrierwave', github: 'carrierwaveuploader/carrierwave'`

mshibuya:
@LIYINGZHEN 's right, it's a master-branch-only feature.

PR #1961, opened 2016-05-25
Define methods instead of alias

daniel-g:
I ran into this issue when trying the multi file uploads feature. When I defined read_uploader and write_uploader, they were not being overriden, so it seems alias_method creates a wire that can't be overriden.


daniel-g:
Forget it, I used undef and define them again in the class.

PR #1963, opened 2016-05-28
0.11 Closing file after mime-type checking

jens-a-e:
Hi!

I was getting `Errno::EACCES:  Permission denied @ sys_fail2 ...` after updating to `0.11.1 or .2` from `0.11.0` on Windows. The added mime-type checking did not close the opened file and Windows is - unfortunately - not happy about it.

I am using `move_to_cache` and `move_to_store` with `true` in my Rails app due to cleanup reasons. Copying the file does of course not cause the error.

Not sure if this is the right place for the issue or if the new mimemagic gem would be better. Regarding the method signature, it is the callers response to close the file though.

Regards,
j

### DIFF
--- a/lib/carrierwave/sanitized_file.rb
+++ b/lib/carrierwave/sanitized_file.rb
@@ -315,9 +315,12 @@ module CarrierWave
     end
 
     def mime_magic_content_type
-      MimeMagic.by_magic(File.open(path)).try(:type) if path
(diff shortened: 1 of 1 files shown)


mshibuya:
Already resolved by #1985.

jens-a-e:
👍

PR #1965, opened 2016-06-02
Fix typo in Google Storage for Developers session

LIYINGZHEN:
When using `config.fog_provider = 'fog-google'`

I encountered an exception said : LoadError: cannot load such file -- fog-google

Solved this with `config.fog_provider = 'fog/google'`


mshibuya:
Thanks!

PR #1970, opened 2016-06-15
README update for Skipping ActiveRecord callbacks

zdennis:
Small doc update as the README to bring the AR callbacks mentioned in line with what it's 0.10 and 0.11.


mshibuya:
What README is saying is about master branch, no need to revert it to the old version.

PR #1972, opened 2016-06-16
Changed public URL method to support empty fog_directory.

### LINKED ISSUE #1971: Fix for Amazon S3 Virtual Hosted Buckets
I'm serving my images from a subdomain of my site, and would like Carrierwave to save images directly into my S3 bucket without uploading into a subdirectory by the same name as my bucket. I'm following this AWS guide on virtual hosted buckets: https://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html

My carrierwave.rb looks like the following with an intentionally omitted config.fog_directory:

```ruby
  config.fog_provider = 'fog/aws'

  config.fog_credentials = {
    provider: 'AWS',
    aws_access_key_id: ENV['S3_ACCESS_KEY_ID'],
    aws_secret_access_key: ENV['S3_SECRET_ACCESS_KEY'],
    region: 'us-east-1',
    host: 'images.example.com',
    endpoint: 'https://images.example.com'
  }
  config.fog_public =  true
  config.fog_attributes = { 'Cache-Control' => "max-age=#{365.day.to_i}" }
end
```

My files upload to the S3 bucket correctly, just the way I want them to. However, Carrierwave will return the wrong public url for the resource with an extra forward slash in the URL after the endpoint. I've fixed the URL issue by making fog_directory optional in the public_url method. I'll be submitting a pull request against this issue in a few minutes.

bentiffany:
This pull request fixes #1971 by avoiding a double forward slash when fog_directory is empty.

### DIFF
--- a/lib/carrierwave/storage/fog.rb
+++ b/lib/carrierwave/storage/fog.rb
@@ -343,7 +343,11 @@ module CarrierWave
             when 'AWS'
               # check if some endpoint is set in fog_credentials
               if @uploader.fog_credentials.has_key?(:endpoint)
-                "#{@uploader.fog_credentials[:endpoint]}/#{@uploader.fog_directory}/#{encoded_path}"
+                if @uploader.fog_directory
+                  "#{@uploader.fog_credentials[:endpoint]}/#{@uploader.fog_directory}/#{encoded_path}"
+                else
+                  "#{@uploader.fog_credentials[:endpoint]}/#{encoded_path}"
+                end
               else
                 protocol = @uploader.fog_use_ssl_for_aws ? "https" : "http"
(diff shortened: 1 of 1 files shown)


bentiffany:
It doesn't appear that this gem receives a lot of attention... If anyone notices this comment and has an idea why tests are failing, I'd love to know which test I need to update to make this PR pass Travis. Also, this was a super critical fix for our project because we were using CNAME based S3 buckets (i.e. images.example.com), and I'd love for others to have access to this fix, too.

mshibuya:
See https://github.com/carrierwaveuploader/carrierwave/issues/1971#issuecomment-241191227.

PR #1976, opened 2016-06-22
Add vi locale

tam-vo:


mshibuya:
Added in [carrierwave-i18n](https://github.com/carrierwaveuploader/carrierwave-i18n/blob/master/rails/locales/vi.yml), thanks!

PR #1979, opened 2016-06-26
Update zh-CN.yml rename whitelist & blacklist

dreamlx:
rename "extension_whitelist_error" to "extension_white_list_error"
rename "extension_blacklist_error" to "extension_black_list_error"


mshibuya:
`extension_whitelist_error` / `extension_blacklist_error` are the correct ones.
cf. https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/locale/en.yml

PR #1936, opened 2016-05-05
Backport fix from PR #1934 to 0.11

locriani:

PR #1985, opened 2016-07-11
Close file in SanitizedFile#mime_magic_content_type

obduk:
Fixes bug where file handles are being left open. This was introduced in #1934
and #1936


mshibuya:
Thanks!

PR #1992, opened 2016-07-29
Adding Arabic locale

AbdullahAs:
Though for _extension_whitelist_error_ for example it will only work if we change it to _extension_white_list_error_. However, following :en locale's labels I kept this one the same.


mshibuya:
Added in [carrierwave-i18n](https://github.com/carrierwaveuploader/carrierwave-i18n/blob/master/rails/locales/ar.yml), thanks!

PR #2013, opened 2016-09-03
Add MIT license badge to the README

kmewhort:
There aren't too many (non-dev/non-test) dependencies in Carrierwave, but I still I ran a scan for compatibility with MIT and all the included gems check out as compatible.

The badge here gives a quick indicator that the license is MIT, and that the libraries used are compliant with that.


mshibuya:
Thanks!

PR #2026, opened 2016-09-21
Add Rails 5.0 stable gemfile for testing.

### LINKED ISSUE #2021: Add Rails 5.0 gemfile
https://github.com/carrierwaveuploader/carrierwave/tree/master/gemfiles

Currently there's no 5.0 gemfile, there's only a rails master gemfile which tests with a version that's very different from 5.0 stable.

Could we have a 50-stable gemfile?

connorshea:
Fixes #2021.

Currently only Rails master is tested, which is now the development version of 5.1. This adds 5.0 to the test matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ gemfile:
   - gemfiles/rails-4-0-stable.gemfile
   - gemfiles/rails-4-1-stable.gemfile
   - gemfiles/rails-4-2-stable.gemfile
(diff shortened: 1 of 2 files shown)


connorshea:
@mshibuya tests are passing now :)

JRuby is not explicitly supported by Rails 5 and so the tests were failing, I've excluded it.

mshibuya:
Merged in, thanks!

PR #2039, opened 2016-10-06
Avoids a GET for Google Cloud storage public_urls

nikz:
Hiya!

As I understand it, the `public_url` method on the Storage class is intended to return a URL without requiring a GET to the backend. At the moment this is how it works for AWS, but for Google Storage the `directory.files.get` call will make an HTTP request.

I've modified this to use the public access URL as [per Google's documentation](https://cloud.google.com/storage/docs/access-public-data). 

I've also added a very simple spec that's similar to the AWS-specific ones. 

Hope this is helpful!

Nik

### DIFF
--- a/lib/carrierwave/storage/fog.rb
+++ b/lib/carrierwave/storage/fog.rb
@@ -355,8 +355,8 @@ module CarrierWave
                 end
               end
             when 'Google'
-              file = directory.files.get(encoded_path)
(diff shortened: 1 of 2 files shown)


mshibuya:
Refs. #1683, 844788609b3c25b0435c42059c9db49d89ba6680

mshibuya:
Thanks 👍

PR #2041, opened 2016-10-10
Remove warnings from specs

cfcosta:
Before:

After:

### DIFF
--- a/spec/support/activerecord.rb
+++ b/spec/support/activerecord.rb
@@ -26,6 +26,6 @@ ActiveRecord::Base.establish_connection(dbconfig.merge(:database => database))
 
 ActiveRecord::Migration.verbose = false
 
-if ActiveRecord::VERSION::STRING >= '4.2'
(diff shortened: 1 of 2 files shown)


spec/support/activerecord.rb
```diff
@@ -25,7 +25,3 @@
 ActiveRecord::Base.establish_connection(dbconfig.merge(:database => database))
 
 ActiveRecord::Migration.verbose = false
-
-if ActiveRecord::VERSION::STRING >= '4.2'
-  ActiveRecord::Base.raise_in_transactional_callbacks = true
```
mshibuya: Without this, ActiveRecord 4.2 produces tons of deprecation warnings.
https://travis-ci.org/carrierwaveuploader/carrierwave/jobs/166306676

cfcosta: Fixed!

mshibuya:
Thanks!

PR #2045, opened 2016-10-18
Add support for callbacks on :proccess

cfcosta:
This PR adds support to running callbacks before and after proccessing a file. The original motivation was to help adding another mitigation tool for exploits that haunt ImageMagick, like this one: https://www.exploit-db.com/exploits/39767/

Some help to writing proper tests would be appreciated, but no other callbacks are tested specifically, and it was tested manually.

### DIFF
--- a/lib/carrierwave/uploader/processing.rb
+++ b/lib/carrierwave/uploader/processing.rb
@@ -71,15 +71,17 @@ module CarrierWave
       def process!(new_file=nil)
         return unless enable_processing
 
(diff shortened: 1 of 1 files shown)


cfcosta:
Hey guys, any input on this? Looks like the build is broken for reasons unrelated to this PR.

cfcosta:
Branch updated, can someone review this?

mshibuya:
OK, merged in.
It would be nice if you could write the specs for callbacks, though.

Issue #1437, opened 2014-08-18
Getting a deprecation warnings  for BeNoLargerThan


torstenb:
Getting the following deprecation warning:

```
CarrierWave::Test::Matchers::BeNoLargerThan implements a legacy RSpec matcher
protocol. For the current protocol you should expose the failure messages
via the `failure_message` and `failure_message_when_negated` methods.
```

jrmyward:
Did this not make it into the `0.11.0` release? I'm still getting the deprecation warnings after updating the gem.

tom-lord:
I'm still getting deprecation warnings, too. Using `0.11.2`.

prrrnd:
Also getting deprecation warnings for `CarrierWave::Test::Matchers::HaveDimensions`, using `0.11.2`

Bastes:
Also getting the deprecation warnings for ```0.11.2```

Bastes:
Hey guys, I've sent a little pull request to allow all of us who aren't really ready to get to a 1.0.0-beta working on their production projects: https://github.com/carrierwaveuploader/carrierwave/pull/2054

PR #2054, opened 2016-10-28
attempt to solve the negative_failure_message rspec warnings

Bastes:
(over v0.11.2 while keeping the negative_failure_message method available for older builds)


Bastes:
Addressing this issue: https://github.com/carrierwaveuploader/carrierwave/issues/1437

Bastes:
Ah, looks like it fails... Although I can't see the logs, and the specs pass on my machine ^^°

mshibuya:
Thanks!

PR #2057, opened 2016-10-31
Don't error if updated record

parterburn:
Fixes issue here: http://stackoverflow.com/questions/21853367/undefined-method-content-length-for-nilnilclass-fog-storage-carrierwave

```
undefined method `content_length' for nil:NilClass
```


mshibuya:
Read the answer there, that's not a valid operation.
http://stackoverflow.com/a/21855579

PR #2074, opened 2016-12-07
mini_magick_image now reads file stream

blaknite:
This fixes not being able to read temporary files during processing a remote file.

Steps to reproduce:

```ruby
class Model < ActiveRecord::Base
  mount_uploader :file, FileUploader
end

class FileUploader < CarrierWave::Uploader::Base
  include CarrierWave::MiniMagick

  process :save_dimensions

  def save_dimensions
    model.file_width = width
    model.file_height = height
  end
end

@model.update_attributes(remote_file_url: 'http://www.example.com/image.png')
# => Errno::ENOENT: No such file or directory @ rb_sysopen - /uploads/tmp/path_to/image.png
```

### DIFF
--- a/lib/carrierwave/processing/mini_magick.rb
+++ b/lib/carrierwave/processing/mini_magick.rb
@@ -317,11 +317,7 @@ module CarrierWave
       end
 
       def mini_magick_image
-        if url
-          ::MiniMagick::Image.open(url)
-        else
-          ::MiniMagick::Image.open(current_path)
-        end
+        ::MiniMagick::Image.read(file.read)
       end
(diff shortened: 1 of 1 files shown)


blaknite:
I spy the Ruby 2.0 error. Will investigate.

janko-m:
The problem with this approach is that you're reading the whole image into memory, which isn't feasible with larger files (the server easily runs out of memory). I'm not sure what a good alternative would be, though.

mshibuya:
Can be risky, unable to merge until a solution comes up.

PR #2078, opened 2016-12-13
Optimization: avoid calling `file.url` twice.

lukeasrodgers:
Since `file.url` does somewhat costly operations, we should avoid calling it twice. More aggressive caching may be possible, but this is an easy and straightforward win.


mshibuya:
Thanks.

Issue #1977, opened 2016-06-23
Invalid locale in CarrierWave::MiniMagick#manipulate!


paweljw:
If a MiniMagick::Error or a MiniMagick::Invalid exception is raised within `manipulate!`, the rescue block will try to find a default error message with

``` ruby
I18n.translate(:"errors.messages.mini_magick_processing_error", e: e, locale: :en)
```

If the :en locale is not included in an application's locales, which is my case, an I18n::InvalidLocale exception will be raised within the rescue block, effectively blocking the developer from gaining insight into what actual exception has occured.

I'm currently using this as a workaround:

``` ruby
rescue ::MiniMagick::Error, ::MiniMagick::Invalid => e
  default = if I18n.available_locales.include?(:en)
              I18n.translate(:"errors.messages.mini_magick_processing_error", e: e, locale: :en)
            else
              ''
            end
  message = I18n.translate(:"errors.messages.mini_magick_processing_error", e: e, default: default)
  raise CarrierWave::ProcessingError, message
end
```

But it's rather specific to my use case and far from perfect for general usage. However, the current implementation is also not perfect in the edge case of apps that do not have a :en locale.

mshibuya:
Dup of #1729, and fixed by #2083.

Issue #1902, opened 2016-03-21
I18n::InvalidLocale error when processing happens to fail and :en is not a valid locale


attilagyorffy:
Hi,

This is an odd one that happens to break my app. I know that ideally the processing (using MiniMagic btw) should not break but in some cases it may actually do and I don't think my app should with the following exception:

``` ruby
I18n::InvalidLocale
```

This happens to be an issue where the MiniMagic processing plugin tries to log an issue and happens to be using the I18n translations with an explicit `:en` locale. Trouble is: :en is not an available locale in my app and so i18n starts to complain (for a valid reason).

Here's what's happening:

```
I18n::InvalidLocale (:en is not a valid locale):
  i18n (0.7.0) lib/i18n.rb:284:in `enforce_available_locales!'
  i18n (0.7.0) lib/i18n.rb:151:in `translate'
  carrierwave (0.10.0) lib/carrierwave/processing/mini_magick.rb:271:in `rescue in manipulate!'
  carrierwave (0.10.0) lib/carrierwave/processing/mini_magick.rb:259:in `manipulate!'
  carrierwave (0.10.0) lib/carrierwave/processing/mini_magick.rb:131:in `resize_to_limit'
  carrierwave (0.10.0) lib/carrierwave/uploader/processing.rb:84:in `block in process!'
  carrierwave (0.10.0) lib/carrierwave/uploader/processing.rb:76:in `each'
  carrierwave (0.10.0) lib/carrierwave/uploader/processing.rb:76:in `process!'
  carrierwave (0.10.0) lib/carrierwave/uploader/callbacks.rb:18:in `block in with_callbacks'
  carrierwave (0.10.0) lib/carrierwave/uploader/callbacks.rb:18:in `each'
  carrierwave (0.10.0) lib/carrierwave/uploader/callbacks.rb:18:in `with_callbacks'
  carrierwave (0.10.0) lib/carrierwave/uploader/cache.rb:122:in `cache!'
  carrierwave (0.10.0) lib/carrierwave/uploader/versions.rb:274:in `block in cache_versions!'
  carrierwave (0.10.0) lib/carrierwave/uploader/versions.rb:259:in `each'
  carrierwave (0.10.0) lib/carrierwave/uploader/versions.rb:259:in `cache_versions!'
  carrierwave (0.10.0) lib/carrierwave/uploader/callbacks.rb:18:in `block in with_callbacks'
[...]

PR #2083, opened 2016-12-16
Fixing #1977 Invalid locale in CarrierWave::MiniMagick#manipulate!

kryzhovnik:
Fixing #1977 Invalid locale in CarrierWave::MiniMagick#manipulate!

Since ([commit on May 9, 2014](https://github.com/svenfuchs/i18n/commit/b5703d7431ecfc1b99409e16c6590448570718ac)) enforce_available_locales in I18n is true by default, the aproach which implies on default error messages from `:en` locale is broked for all apps where `:en` is not available locale. It's very suddenly when image processing raise `I18n::InvalidLocale` error.

In this pull request I propose to remove the "default fallback" and let carrierwave users decide themselves how to deal with missed locales and/or translations.

### DIFF
--- a/lib/carrierwave/processing/mini_magick.rb
+++ b/lib/carrierwave/processing/mini_magick.rb
@@ -303,8 +303,7 @@ module CarrierWave
         image.destroy!
       end
     rescue ::MiniMagick::Error, ::MiniMagick::Invalid => e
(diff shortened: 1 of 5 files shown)


kryzhovnik:
Same issue #1902

mshibuya:
Agreed, thanks!

PR #2091, opened 2017-01-01
Fix args of `default_url` in templates

znz:
README says `def default_url(*args)` instead of `def default_url`


mshibuya:
Thanks.

Issue #1889, opened 2016-03-04
Uploader::Cache#cache_id= bug


Okasu:
https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/uploader/cache.rb#L193
This one will fail if time is a negative number. Well, that may seem unimportant but look at recent iOS bricking bug related to Unix time...

mshibuya:
Fixed by #2092.

PR #2092, opened 2017-01-02
Fixing #1889 Cache ID should not fail from negative time stamp

NickOttrando:
Fix for #1889 adjusted regex to allow for negative time stamps.


mshibuya:
Merged in, thanks!

PR #2097, opened 2017-01-13
Allow non-argument options to be passed in mini magick combine_options

krismartin:
When a non-argument option is passed to mini magick processing using combine_options, the processing fails.

For example:

```
resize_to_fill(320, 320, combine_options: {strip: nil})
```

generates this command:

```
mogrify -resize 320 -gravity Center -background rgba(255,255,255,0.0) -strip  /var/folders/1_/tdx0fnfx32g1hqv94s3ttkw00000gn/T/mini_magick20170113-86521-eezfq7.jpg
```

But processing is failing with mini_magick_processing_error. Changing `strip: nil` to `strip: ""` (empty string) also did not solve the problem.

The only way I can get it to work is by changing the [append_combine_options](https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/processing/mini_magick.rb#L313) method to invoke the method without any arguments if nil options is passed:

```
def append_combine_options(cmd, combine_options)
    combine_options.each do |method, options|
        if options.nil?
            cmd.send(method)
        else
            cmd.send(method, options)
        end
    end
end
```

### DIFF
--- a/lib/carrierwave/processing/mini_magick.rb
+++ b/lib/carrierwave/processing/mini_magick.rb
@@ -312,7 +312,11 @@ module CarrierWave
 
       def append_combine_options(cmd, combine_options)
         combine_options.each do |method, options|
-          cmd.send(method, options)
+          if options.nil?
+            cmd.send(method)
+          else
+            cmd.send(method, options)
+          end
         end
       end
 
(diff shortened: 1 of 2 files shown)


mshibuya:
Sounds good, thanks!

PR #2108, opened 2017-01-31
Added fog_aws_accelerate options for fog storage provider

krekoten:
AWS S3 supports accelerated file transfers to and from S3.
[Transfer Acceleration](http://amzn.to/2jQmpAX)

This introduces ability to configure Carrierwave to return
accelerated public urls since Carrierwave constructs them
by itself for speed concerns.

To enable this feauture you need to set fog_aws_accelerate
to true. It is false by default.

Changes added:

* fog_aws_accelerate added to fog aws configuration options
* corresponding specs updated
* corresponding docs updated


krekoten:
Hm, it seems build fails due to bundler cannot resolve some gems. That is interesting because Gemfile and lock files were not changed.

atz:
Same blocker on CI for my PR.  Looks like some underlying dependency maintenance to perform.

mshibuya:
Looks nice, thanks!

PR #2051, opened 2016-10-27
[skip ci] Wip on downloader strategies

ym97:

PR #2110, opened 2017-02-01
Allow carrierwave to do safe or all redirections based on configuration

zdraganov:
I see plenty of historical issues to implement this and the best approach I see is [this PR](https://github.com/carrierwaveuploader/carrierwave/pull/2051) which seems to be opened in 2015. 

I've implemented my idea of option about download safety, so if you like it you can merge.

All the best from snowy Bulgaria folks! Keep doing such good work 😄

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,9 @@ spec/test.log
 spec/tmp
 *.swp
 .rvmrc
+.ruby-version
 .idea
 .bundle
 Gemfile.lock
 gemfiles/*.lock
+
--- a/carrierwave.gemspec
+++ b/carrierwave.gemspec
@@ -42,4 +42,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency "timecop", "0.7.1"
   s.add_development_dependency "generator_spec", ">= 0.9.1"
   s.add_development_dependency "pry"
+  s.add_development_dependency "open_uri_redirections"
 end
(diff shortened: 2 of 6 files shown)


janko-m:
Hmm, isn't the `open-uri` option for disabling/enabling redirects called `:redirect` instead of `:allow_redirections`?

zdraganov:
Yeah, they are. But to be fully functional I wanted to implement http->https redirections with `:safe` option, and additionally https->http redirection with `:all` configuration option. That is achieved using external dependency called [open-uri-redirections](https://github.com/open-uri-redirections/open_uri_redirections) which is simply patching the redirection logic and adding thise functionality.
IMO same full functionality could be achieved with this idea https://github.com/carrierwaveuploader/carrierwave/pull/2051/files, but as I see it's not tested at all and there is a `puts` in the code.

mshibuya:
This appears to be overkill, isn't it enough to require `open_uri_redirections` in your app and set `uploader.remote_image_request_header =  {allow_redirections: :all}` ?

zdraganov:
I've didn't notice that option. I will test it and will give feedback.

zdraganov:
@mshibuya where I can find `remote_image_request_header` option documented?

zdraganov:
Just found that this option is missing in our current version (0.11.2), but existing in the latest carrierwave versions.
Closing it for now.

PR #2112, opened 2017-02-02
Allow read to accept bytes as parameter

hexgnu:
Technically you should be able to call #read(bytes) with IO objects for
things like Zlib::GzipReader etc since they are chunked.

I have been able to do that with files that are locally although I doubt
that it would ever work with Fog just how it's built. This at least gets
over the issue I found related to the following code snippet.

```ruby
page = Page.first
page.carrierwave_obj = File.open('./file.gz', 'rb')
page.save!

Zlib::GzipReader.new(page.carrierwave_obj)
```

Thanks for reading this. Please let me know if there's anything else I can do to assist with this.

### DIFF
--- a/lib/carrierwave/storage/fog.rb
+++ b/lib/carrierwave/storage/fog.rb
@@ -273,7 +273,7 @@ module CarrierWave
         # === Returns
         #
         # [String] contents of file
-        def read
+        def read(bytes = nil)
           file.body
         end
 
--- a/lib/carrierwave/sanitized_file.rb
+++ b/lib/carrierwave/sanitized_file.rb
@@ -156,14 +156,14 @@ module CarrierWave
     #
     # [String] contents of the file
     #
-    def read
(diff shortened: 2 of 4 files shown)


lib/carrierwave/storage/fog.rb
```diff
@@ -273,7 +273,7 @@ def initialize(uploader, base, path)
         # === Returns
         #
         # [String] contents of file
-        def read
+        def read(bytes = nil)
           file.body
```
mshibuya: This returns whole content, is it really OK?

hexgnu: When I tried it it seemed to be fine to return the entire body. I couldn't figure out a way of returning chunks in fog

mshibuya:
Sorry for silence, but I've decided not to merge this because I'm not sure if it is safe to return whole content when bytes is given. It will depend on implementation and might break caller's expectation.

Issue #1841, opened 2016-01-19
Content Type is lost when using model instance


rsiddle:
I'm uploading a video MP4 file and by default it's returning the content_type as "application/mp4" instead of "video/mp4".

At the point below within the uploader, the pry statement is showing the correct content_type

```
process :set_content_type

  def set_content_type(*args)
    content_type = file.content_type == 'application/octet-stream' || file.content_type.blank? ? MIME::Types.type_for(original_filename).first : file.content_type

    self.file.instance_variable_set(:@content_type, content_type)
    binding.pry
  end
```

Here is the output at this point
`
pry(#<AttachmentUploader>)> self.file.content_type
=> "video/mp4"
`

in my upload.rb file, I use

```
attachment.content_type
```

The result is 

```
pry(#<Upload>)> attachment.content_type
=> "application/mp4"
```

I can't figure out why CarrierWave is changing the content_type

mehlah:
Thanks @rsiddle for the report.
Which version of CW are you using?

dsandstrom:
This happens with 1.0.0.beta

olegantonyan:
This bug appeared to me after update to 1.0.0

ngoral:
1.2.2 is also affected

mshibuya:
#2117 should work, please provide detail if you're still having this issue.

PR #2117, opened 2017-02-08
Fix content type being reset when using file cache storage

dweinand:
There has been a longstanding issue in `SanitizedFile#move_to` where an existing content type would be lost. This was exacerbated by the changes in #1312, which made the file cache store use `move_to` all the time.

Prior to this fix, anyone using the file cache store will lose any explicitly set content type. This can be seen in: #1841.

This PR solves the underlying `SanitizedFile` issue, and adds some specs to ensure the same issue isn't present in the fog storage.


Tom-Tom:
Hi, any news regarding this PR ?
Thx

mshibuya:
Looks good, thanks!

Issue #2116, opened 2017-02-07
SanitizedFile is not URI-aware, e.g. misinterprets S3 filename


atz:
```ruby
z = CarrierWave::SanitizedFile.new('https://me.s3.amazonaws.com/path/to/image.png?X-Amz-Security-Token=FQoDYXdzEMT//////////whatever')
 => #<CarrierWave::SanitizedFile:0x007f9b51221458 @file="https://me.s3.amazonaws.com/path/to/image.png?X-Amz-Security-Token=FQoDYXdzEMT//////////whatever", @original_filename=nil, @content_type=nil> 
z.original_filename
 => "whatever" 
z.filename
 => "whatever"
```

Here we would expect `image.png`.  

`sanitize` would be improved in both security and utility by using dedicated a URI parsing lib in addition to `File` before regex, like:
```ruby
File.basename(URI.parse(z.file).path)
 => "image.png"
```

What is remarkable is that we otherwise have the gem working in our rails models and AWS setup for the same kind of input, but the filename corruption causes problems in the UI downstream.  If this is an acceptable course, I can cook up a PR.

atz:
OK, I see that there is some tension inasmuch as `URI` would throw `URI::InvalidURIError: bad URI(is not URI?)` for many of the strings that otherwise get sanitized.  But that can still be accommodated, IMHO.

PR #2118, opened 2017-02-08
Allow some URI-compatibility, reveal latent bug w/ tests

atz:
See: https://github.com/carrierwaveuploader/carrierwave/issues/2116

- Use explicit subject w/ 1-liner `it` blocks
- Don't stub, use a real object.

In particular `original_filename` has enough complexity that it is unclear that the behavior of stubbed objects corresponds to instances that are actually possible to create in the wild.

**Case in point**, the nix directory test no longer passes, though it is ostensibly doing the same thing!

```
expected: "_."
     got: "carrierwave"
```
That is a result of, effectively, `File.basename(File.expand_path(@file))` between `original_filename` and `path`.


atz:
Current CI failures are from dependency issues in a couple cells of your massive build matrix.

mshibuya:
Thank you and sorry for taking so long, staged for 2.0.0.

mshibuya:
After careful consideration while working on #2119, I've come to revert this, because what SanitizedFile aims to do is provide safer way to handle **local** files, not for parsing URLs for filename extraction or treating untrusted paths.

PR #2130, opened 2017-03-02
Fix Rails 5.1.0 deprecation warnings

paulsturgess:
Firstly thanks for a great gem! This is my first PR, so hopefully I'm on the right lines here.

Rails 5.1.0 introduces deprecation warnings for methods around dirty attributes in ActiveRecord as outlined here: https://github.com/rails/rails/commit/16ae3db5a5c6a08383b974ae6c96faac5b4a3c81

This PR fixes the deprecation warnings by switching the dirty attribute method used, based on the ActiveRecord version in use.

### DIFF
--- a/lib/carrierwave/mount.rb
+++ b/lib/carrierwave/mount.rb
@@ -186,7 +186,8 @@ module CarrierWave
         end
 
         def store_previous_changes_for_#{column}
-          @_previous_changes_for_#{column} = changes[_mounter(:#{column}).serialization_column]
+          attribute_changes = ::ActiveRecord.version.to_s.to_f >= 5.1 ? saved_changes : changes
+          @_previous_changes_for_#{column} = attribute_changes[_mounter(:#{column}).serialization_column]
         end
 
         def remove_previously_stored_#{column}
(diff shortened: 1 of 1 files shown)


paulsturgess:
Looking at the build failures on Travis it looks like older versions of Ruby aren't too happy at installing some of the gems `rack (2.0.1)` and `nio4r (2.0.0)`.

Locally for me the specs all pass (except the one pending one) running either ruby 2.4.0 or 2.3.3p22

.travis.yml
```diff
@@ -22,6 +22,7 @@ gemfile:
   - gemfiles/rails-4-1-stable.gemfile
   - gemfiles/rails-4-2-stable.gemfile
   - gemfiles/rails-5-0-stable.gemfile
+  - gemfiles/rails-5-1-0-beta1.gemfile
```
mshibuya: This is not to be added. We're already testing Rails master branch, where Rails 5.1.0 development is in progress.

paulsturgess: OK no problem. Removed. 👍

mshibuya:
Perfect, thanks!

paulsturgess:
Glad I could help. Out of interest the master branch run on Travis was failing the spec:
`CarrierWave::ActiveRecord#mount_uploaders #save should remove the images if remove_images? returns true`

and also on the spec:
`CarrierWave::ActiveRecord#mount_uploaders remove_images! should return to false after being saved`

Do these matter or is that unrelated?

PR #2139, opened 2017-03-13
Link to CarrierWave questions on Stack Overflow

jish:
I clicked on the link in the readme and got really confused when I saw a bunch of questions about ASP.net. I propose that the readme links to the `carrierwave` tag on Stack Overflow.

Thoughts?


mshibuya:
❤️

PR #2140, opened 2017-03-14
Remove potentially redundant HEAD request

eritiro:
When you check if a file exists using Fog, you can potentially end up with an extra HEAD request to the remote server: one request to check its existence and another to read the file.

Consider the following code:
```
if fog_file.exists?
  fog_file.read
end
```
It will produce two HEAD request and one GET
With this PR, it will produce only one HEAD and one GET.


mshibuya:
Agreed, thanks.

PR #2146, opened 2017-03-24
Update CHANGELOG.md

earksiinni:
1.0.0 changed the method signature for `#default_url`, which breaks old uploaders' overridden `#default_url` implementations that don't allow for any arguments to be passed. A quick fix for users is to change any instances `def default_url` to `def default_url(*args)`.


mshibuya:
Thanks.

PR #2151, opened 2017-03-31
Correct syntax in README

fedorkk:
After a lot of attempts, only this example works for me.
Syntax is taken from fog: https://github.com/fog/fog-google/blob/db446d1d8fbf0b75e13b4c39af3eb6b1ecb3101c/lib/fog/storage/google_json/models/file.rb
and tested for fog-google, and fog-aws.


mshibuya:
👍

PR #2162, opened 2017-04-20
Add fog_serve_public as an override

rnaud:
## What
For security purposes, having in a bucket a mix of files that are meant to be public and others that are meant to be PII data is a security breach waiting to happen.

This PR allows to decouple fog_public from fog_serve_public. Which means you can upload files as private but still serve them as public files through Cloudfront or through your provider.

We want to do this so that we can remove the permission for the user to set ACL on a per file basis but still allows for public access to these files.


rnaud:
Closing the PR, it turns out you can just set `fog_attributes = {public: false}` to enforce us never setting ACL on any file.

Issue #2167, opened 2017-05-01
Resize to Fill with Dimensions from Database


tomprats:
I would like to store dimensions elsewhere in the database so that they can be configured from the app. When updated, it will trigger a background job to resize all the current images for that uploader. When a new image is I want it to use those dimensions

Right now I'm stuck on how to change

`process resize_to_fill: [50, 50]`

to something like

`process resize_to_fill: [Setting.find_by(uploader: "Fun").width, 50]`

Let me know if you have any ideas.

Thanks,
Tom

tomprats:
Fixed in #2169

Issue #464, opened 2011-09-18
Dynamic versions


grex84:
Hi,
I'm looking for solution to be able resize images depending on
my model settings.

Something like that would be what I'm looking for, but it is not
working :(

version :thumb do
   process :resize_to_limit => model.thumb_size
end

I know you can access model setting when setting the store_dir but not for versions.  
Is this a bug or is there a way to do this?

Your help would be greatly appreciated!

PR #2169, opened 2017-05-06
Added Proc Support for Width and Height

tomprats:
I opened this issue (https://github.com/carrierwaveuploader/carrierwave/issues/2167) earlier this week and found a couple hacky solutions, so I thought I'd offer a more direct solution.

I then found this issue (https://github.com/carrierwaveuploader/carrierwave/issues/464) today and I think my fix would solve that problem as well, but may need to be updated to include something like [this](https://github.com/carrierwaveuploader/carrierwave/blob/540a84c606353a878d6574740e44f7970b754463/lib/carrierwave/uploader/configuration.rb#L150) in case some context needs to be passed in.

Let me know what you think! Thanks,
Tom

### DIFF
--- a/lib/carrierwave/processing/mini_magick.rb
+++ b/lib/carrierwave/processing/mini_magick.rb
@@ -127,6 +127,8 @@ module CarrierWave
     # [MiniMagick::Image] additional manipulations to perform
     #
     def resize_to_limit(width, height, combine_options: {})
+      width = width.call if width.instance_of?(Proc)
+      height = height.call if height.instance_of?(Proc)
       manipulate! do |img|
         img.combine_options do |cmd|
           cmd.resize "#{width}x#{height}>"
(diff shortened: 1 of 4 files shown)


tomprats:
Holy cow this actually works! Using my PR's code, I can now look up resizing values based on a field on the current model (or even the image being uploaded)

```
class SettingUploader < CarrierWave::Uploader::Base
  def self.width
    Proc.new do |uploader|
      # Default to enormous value to ensure height is used
      uploader.setting["width"] || 1000000000
    end
  end

  def self.height
    Proc.new do |uploader|
      # Default to enormous value to ensure width is used
      uploader.setting["height"] || 1000000000
    end
  end

  process resize_to_fit: [width, height]

  def setting
    # Get dimensions from DB or default to the image's current dimensions
    Setting.find_by(name: model.name) || { "width" => width, "height" => height }
  end
end
```

Hopefully someone decides to look at this at some point. For me, this feature adds a ton of capabilities for CarrierWave

tomprats:
@mshibuya or whoever: Can I get your thoughts on this? Is there a better person to ping to get this moving?

mshibuya:
Thanks for the PR, it will be nice to have a spec against a Proc with arity = 1(takes argument).

tomprats:
@mshibuya I updated the tests and pulled in master to ensure other tests still pass

mshibuya:
Thanks! 👍

PR #2178, opened 2017-05-17
Replace send with __send__

dminuoso:
Right now if a user overrides the `send` method in their model it breaks the mount_upload helper and carrierwave validations. By using `__send__` instead it avoids possible name clashes for when the user does override `send` in their model.


mshibuya:
Thanks.

PR #2185, opened 2017-05-29
Fix issue encountered when trying to reopen a file stored  remotely

zog:
When one tries to reopen a fog-stored file after it has been persisted, they get an error:

>  Magick::ImageMagickError: unable to open image

This is because the RMagick processor tries to open the file from its `current_path`, which in this case is the path on the remote server.
This is easily fixed by letting the uploader get the content og the image and send it to `  ::Magick::Image.from_blob`


thiagofm:
Can you please add tests?

zog:
There you are.

spec/processing/rmagick_spec.rb
```diff
@@ -220,6 +220,38 @@
     end
   end
 
+  describe "#rmagick_image", focus: true do
```
mshibuya: `focus: true` needs to be removed.

zog: Oops :)

mshibuya:
Nice, thanks!

PR #2191, opened 2017-06-02
Check if files are identical via FS rather than name before move

riffraff:
in `#move_to`  we rely on `FileUtils.mv` which causes an error if trying to move a file to itself.

There is a check to prevent this which tries to check if the file is the same by looking at the path, but this is not enough on systems which have a case insensitive file system (OSX/macOS with HFS+).

If we use the same check that `FileUtils` would do, i.e. rely on the file system itself, the problem should be solved. 

I am not sure how to write a test for this though, as it would generally pass on platforms with more reasonable file systems.


mshibuya:
Added spec will fail on case-sensitive filesystems(including Travis), right?

riffraff:
I thought on a case-sensitive FS it would pass both with and without the fix, cause the test only checks that no error is raised, like this other one 
https://github.com/riffraff/carrierwave/blob/841ccab7109bfba77e15ee9fa937bcc25ddde2e3/spec/sanitized_file_spec.rb#L408

But it seems there is a problem because a `before`  block in the including context checks the `sanitized_file`  let variable when the target file has not been created yet. 

Sorry, I tried running this on linux but I'm getting a host of other failures like
`undefined method `current_path' for #<CarrierWave::SanitizedFile:0x005612ea7fc558>`
which makes it hard to tell if any fix works there. 

I've changed the spec slightly so that this should pass on linux, by using a differently named file. 
This is not ideal cause it does one useless setup though, but I am not sure how to improve it.

thiagofm:
I think it's good as it is. Thanks a lot for your effort 👍

PR #2198, opened 2017-06-15
Modify README.md to add iam_profile credentials configuration

alejandroperea:
Instead of using AWS secret and access key, we can also use IAM Profile credentials.

### DIFF
--- a/README.md
+++ b/README.md
@@ -671,6 +671,16 @@ CarrierWave.configure do |config|
 end
 ```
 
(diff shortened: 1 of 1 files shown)


thiagofm:
This is mostly related and very specific to fog. I guess somebody trying to do that would check fog out, for that reason I would like to not merge this, unless IAM roles becomes a big trend and more people ask for it.

Thanks a lot for your effort, though!

PR #2199, opened 2017-06-19
Change file size of error message to human size

aki77:
**before**
`File size should be less than 1234567`

**after**
`File size should be less than 1.18 MB`


thiagofm:
Can you write a test for it?

Thanls.

aki77:
Done.

mshibuya:
Nice, thanks!

aki77:
Thanks!

PR #2207, opened 2017-07-23
Added resize_to_fit_with_blurred_bg to MiniMagic

scr1pt:
I wonder why this feature still isn't implemented.


mshibuya:
This feature is too specific for your use-case, my recommendation is to create a how-to entry in the [Wiki](https://github.com/carrierwaveuploader/carrierwave/wiki#how-tos).

PR #2209, opened 2017-07-26
Manage transaction rollback

skosh:
When we try to create new active record object into transaction and meet rollback:
    database row will not be created but uploader has created file in folder and doesn't remove it.

  I changed behavior: now we store file only after commit action. This true and for update action in transaction.

  Changes in spec/orm/activerecord_spec.rb:
    - Add tests which covers this flows
    - replace cleaning folder from 'file_path' to 'public_path'. Cause file_path is 'spec/fixtures'
      and we don't create any 'uploads' folders there. We works with 'spec/public/uploads' folder.

### DIFF
--- a/lib/carrierwave/orm/activerecord.rb
+++ b/lib/carrierwave/orm/activerecord.rb
@@ -52,13 +52,12 @@ module CarrierWave
       validates_processing_of column if uploader_option(column.to_sym, :validate_processing)
       validates_download_of column if uploader_option(column.to_sym, :validate_download)
 
-      after_save :"store_#{column}!"
       before_save :"write_#{column}_identifier"
+      after_save :"store_previous_changes_for_#{column}"
       after_commit :"remove_#{column}!", :on => :destroy
       after_commit :"mark_remove_#{column}_false", :on => :update
-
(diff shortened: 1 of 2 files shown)


oakypinky:
👍

sashakovryga:
👍

mshibuya:
Great work, this will be included in 2.0.0 release!

fsateler:
I believe this change should be reverted. I'm not saying it doesn't fix the issue it fixes, but it introduces other problems:

- Uploader state is inconsistent between `save` and `commit`. For example, reading `my_object.file_url.url` will return a bogus url if I read before commit.
- A file might not be uploaded after a successful commit. For example, if there is a network failure. This causes the transaction to not be rolled back and invalid records being recorded.
- Even if the upload eventually completes, it can take an arbitrary amount of time, during which the record is invalid.

Thus, this change can introduce data loss and data corruption.

I believe a correct fix for the original issue is to add an `after_rollback` callback that (tries to) removes the destination files.

mshibuya:
@fsateler Makes sense, could you open a new issue for that? Or even better if you could work on such an implementation.

fsateler:
Sure, done at #2545

PR #2223, opened 2017-09-01
add require statement

jasdeepsingh:
otherwise raises: `uninitialized constant CarrierWave::Utilities::Uri::URI (NameError)` on a dependent project: https://github.com/metaware/carrierwave-google-storage


mshibuya:
Thanks!

Issue #2173, opened 2017-05-08
translation missing: en.errors.messages.extension_whitelist_error


jdzak:
I'm using `carrierwave (1.1.0)` and `rails (5.0.2)` and i'm getting `translation missing: en.errors.messages.extension_whitelist_error` when I try an upload a file that isn't in the whitelist. I looked at the i18n load path and don't see carrierwave in the load path. Any ideas?

My uploader has:
```
class UserProfilePhotoUploader < CarrierWave::Uploader::Base
  def extension_whitelist
    %w[jpg jpeg gif png]
  end
end
```

jdzak:
So I just stumbled upon [this discussion](https://github.com/carrierwaveuploader/carrierwave/issues/1823) about carrierwave-i18n and a corresponding [PR](https://github.com/carrierwaveuploader/carrierwave/pull/1848).

Seems to me the intent was to keep `carrierwave/locale/en.yml` in the `carrierwave` repo and pull all the other localizations into `carrierwave-i18n` repo. It seems the PR did that, but it also [removed part of the initializer that includes carrierwave in i18n load path](https://github.com/carrierwaveuploader/carrierwave/pull/1848/files#diff-b9b95ffaa8fe82aef6a5d922757578dbL50) making the `carrierwave/locale/en.yml` file unloadable.

Should the code to add the carrierwave localizations directory to the i18n load path be re-added?

leoi11:
Running into the exact same error. Looks right to me that we need to add it back..

leoi11:
Don't have time to send a PR yet, but currently I'm deploying a VERY HACKY workaround using Bundler (>1.3) (please refrain to use):
In config/initializers/carrierwave.rb:
```
carrierwave_gem = Bundler.rubygems.find_name('carrierwave').first
I18n.load_path.prepend(File.join(carrierwave_gem.full_gem_path, 'lib', 'carrierwave', 'locale', "en.yml")) if carrierwave_gem.present?
```

johnfoley3:
For anyone else running into this issue, the better answer is to add
```
gem 'carrierwave-i18n'
```
to your gem file

PR #2226, opened 2017-09-12
Update locales to current master

tom-lord:
Also added a reference to the associated `carrierwave-i18n` library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -901,18 +901,27 @@ mount_uploader :avatar, AvatarUploader, mount_on: :avatar_file_name
 
 ## I18n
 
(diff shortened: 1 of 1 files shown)


tom-lord:
The current keys are defined here: https://github.com/carrierwaveuploader/carrierwave-i18n/blob/master/rails/locales/en.yml

(Should the README reference this file, too? I wasn't sure.)

This PR is in response to various out standing issues about "missing translations", such as https://github.com/carrierwaveuploader/carrierwave/issues/2173

mshibuya:
Thanks 👍

PR #2230, opened 2017-09-21
Reducing allocations while reading configuration

yosiat:
Instead of passing string to `instance_variable_defined?`, we pass a symbol, which saves
an allocation and on micro benchmarks it's faster.


yosiat:
@mshibuya  @thiagofm can you please look at this PR?

mshibuya:
Thanks!

yosiat:
@mshibuya do you plan to release version?

PR #2236, opened 2017-10-06
Corrected explanation of option resize_to_fit

jwipeout:
- more robust explanation of resizing images.
- explained alternative option of resize_to_fill which crops.


mshibuya:
👍

PR #2237, opened 2017-10-09
Reset Content-Type when converting a file format

kyoshidajp:
Hi.

The original Content-Type is preserved, although the file format has been converted. 
It is a problem in my application, which returns the Content-Type by extracting from the stored file.

I would be happy if the Content-Type is set again with a converted new file.

### DIFF
--- a/lib/carrierwave/sanitized_file.rb
+++ b/lib/carrierwave/sanitized_file.rb
@@ -185,6 +185,7 @@ module CarrierWave
       mkdir!(new_path, directory_permissions)
       move!(new_path)
       chmod!(new_path, permissions)
+      self.content_type = ::MIME::Types.type_for(new_path).first.to_s
       if keep_filename
         self.file = {:tempfile => new_path, :filename => original_filename, :content_type => content_type}
       else
--- a/lib/carrierwave/sanitized_file.rb
+++ b/lib/carrierwave/sanitized_file.rb
@@ -185,7 +185,6 @@ module CarrierWave
       mkdir!(new_path, directory_permissions)
       move!(new_path)
       chmod!(new_path, permissions)
-      self.content_type = ::MIME::Types.type_for(new_path).first.to_s
       if keep_filename
         self.file = {:tempfile => new_path, :filename => original_filename, :content_type => content_type}
       else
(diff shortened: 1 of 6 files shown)


krnjn:
+1

raychatter:
Good find! This is a problem for some of our code, too. We are using `CarrierWave::RMagick`:
```ruby
version :pdf do
  process convert: 'pdf'

  def full_filename(for_file = model.file.file)
    super.chomp(File.extname(super)) + '.pdf'
  end
end
```

The filename was getting changed properly, but the content type was getting set to `image/png`, which was the content type of the original file. Thus, the PDF file could not be opened.

With the changes in this PR, the content type is being properly set to `application/pdf`.

kyoshidajp:
@mshibuya Could you review this PR 🙇

lib/carrierwave/sanitized_file.rb
```diff
@@ -185,6 +185,7 @@ def move_to(new_path, permissions=nil, directory_permissions=nil, keep_filename=
       mkdir!(new_path, directory_permissions)
       move!(new_path)
       chmod!(new_path, permissions)
+      self.content_type = ::MIME::Types.type_for(new_path).first.to_s
```
mshibuya: Setting content_type in `move_to` is not intuitive, just moving file usually doesn't cause content-type change.

How about setting content_type in `manipulate!` ?
https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/processing/mini_magick.rb#L295
https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/processing/rmagick.rb#L341

kyoshidajp:
I have moved it.

mshibuya:
Great, thanks!

raychatter:
@mshibuya would you mind doing another release so that we can get this change as an official version in our Gemfile (instead of having to rely on the commit)? Our team is converting file formats and it would help us out to have an official version. Thanks! 😃

krnjn:
@mshibuya sorry to bother, but would you be able to create a new release with this merged code? We would like to use the official release if possible rather than the master branch / commit hash. Thanks!

mshibuya:
Just released 1.2.2 ✨

raychatter:
@mshibuya Thank you so much!

PR #2270, opened 2018-01-16
Add allowed_types to content_type_whitelist_error

mhluska:
We already do this in `extension_whitelist.rb` so this just matches the functionality.


mshibuya:
Looks good, but we need a test case which covers this feature.

mhluska:
@mshibuya @jedrekdomanski I added the test now.

jedrekdomanski:
Looks good to me

mshibuya:
Great, thanks!

PR #2274, opened 2018-01-29
Add `use_iam_profile` aws config parameter

thepoho:
Good AWS practice dictates we should be using instance profiles where possible for AWS s3 permissions (provided we are all within AWS land).  I had to go digging to find this parameter, so upgraded the documentation to make it easier for anyone looking for this in future.


mshibuya:
Thanks!

PR #2292, opened 2018-03-11
Replace mime-types gem with mini_mime

bradleypriest:
Replaces the MimeTypes gem with MiniMime.

Similar to https://github.com/mikel/mail/pull/1059 and https://github.com/teamcapybara/capybara/pull/1884, this switches out the Mime Types gem with [MiniMime](https://github.com/discourse/mini_mime) to save memory.

EDIT:
There seems to be some requirements around the integration between carrierwave and fog going on here.

I see a couple of options:
1) We more explicitly call out that if they're using fog, they'll need to include the `mime-types` gem.
2) We try and convince `fog` to move across to `mini_mime` as well.
3) We add a mime-helper module that defines the three methods we use the mime gems for which can delegate to whichever library is included in the Gemfile.
4) We don't support mini_mime

With the incredibly popular [mail](https://github.com/mikel/mail/) and [capybara](https://github.com/teamcapybara/capybara) gems moving to mini_mime, it'd be nice to switch carrierwave across too (or at least provide the option)


bradleypriest:
This PR has been open a year now, I appreciate it's set up for the 2.0 release of the gem, which is an absolutely herculean task.

However for my own mental wellbeing (I can't deal with a PR being open this long), I'm going to close it now.

Anyone is welcome to this code at any time, whether it's by reopening this PR, or recreating a new one.

Thanks for the hard work.

mshibuya:
Sorry for being late, we appreciate your work!

PR #2297, opened 2018-04-03
When the file pointed to by a CarrierWave::Storage::Fog::File ...

j3pic:
When the file pointed to by a `CarrierWave::Storage::Fog::File` does not exist, throw a `CarrierWave::IntegrityError` when the private `file` method is called.

Otherwise, if you call `read` on the object, a `NoMethodError` is thrown because `file` ends up returning `nil`, and then `read` attempts to call `nil.body`.

```
Error: undefined method `body' for nil:NilClass (Error of type NoMethodError) /var/app/20180403T073104Z/vendor/bundle/ruby/2.3.0/gems/carrierwave-0.11.2/lib/carrierwave/storage/fog.rb:226:in `read' /var/app/20180403T073104Z/vendor/bundle/ruby/2.3.0/gems/carrierwave-0.11.2/lib/carrierwave/uploader/cache.rb:89:in `sanitized_file' /var/app/20180403T073104Z/vendor/bundle/ruby/2.3.0/gems/carrierwave-0.11.2/lib/carrierwave/uploader/cache.rb:128:in `cache!' /var/app/20180403T073104Z/vendor/bundle/ruby/2.3.0/gems/carrierwave-0.11.2/lib/carrierwave/uploader/cache.rb:85:in `cache_stored_file!' 
```

### DIFF
--- a/lib/carrierwave/locale/en.yml
+++ b/lib/carrierwave/locale/en.yml
@@ -6,6 +6,7 @@ en:
       carrierwave_download_error: could not be downloaded
       extension_whitelist_error: "You are not allowed to upload %{extension} files, allowed types: %{allowed_types}"
       extension_blacklist_error: "You are not allowed to upload %{extension} files, prohibited types: %{prohibited_types}"
+      fog_file_not_found_error: "File not found in remote directory: %{path}"
       content_type_whitelist_error: "You are not allowed to upload %{content_type} files"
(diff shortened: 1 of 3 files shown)


j3pic:
Hmmm. Looks like your library tries to avoid throwing exceptions when files don't exist remotely.

PR #2298, opened 2018-04-05
Delegate MiniMagick processing to ImageProcessing gem

### LINKED ISSUE #2062: Performance drop of MiniMagick::Image.open vs. MiniMagick::Image.new
My use case is that I want to create image previews from the first page of multi-page PDF document uploads.

I've used @davidchristiandy's solution [from here](https://github.com/carrierwaveuploader/carrierwave/pull/691#issuecomment-186332089).

```ruby
version :cover do
  process :convert_cover => :png
end

def convert_cover(format)
  manipulate! do |img| # this is ::MiniMagick::Image instance
    img.format(format.to_s.downcase, 0)
    img
  end
end
```

The problem I noticed is a significant performance drop for large PDFs like books with tens/hundreds of pages which didn't make sense to me. Uploading one such book took about **9 seconds** to process. So, I began tracing the problem.

First, I converted the first page directly using `imagemagick` and it ran in less than **1 second**.
```
$ convert book.pdf[0] test-imagemagick.png
```

Then, I used `MiniMagick` directly from a `rails console` and it also ran in less than **1 second**.

```ruby
img = MiniMagick::Image.new('book.pdf') # takes about 1 second
img.format('png', 0)
```

Looking through the implementation i found that `manipulate!` uses [MiniMagick::Image.open](https://github.com/carrierwaveuploader/carrierwave/blob/v0.11.2/lib/carrierwave/processing/mini_magick.rb#L260).

When I ran that in `rails console`, I had found the performance drop.

```ruby
img = MiniMagick::Image.open('book.pdf') # takes about 9 seconds
```

Turns out, the [MiniMagick::Image.open](http://www.rubydoc.info/github/minimagick/minimagick/MiniMagick%2FImage.open) method makes a copy of the loaded file instead of manipulating it directly and that's most likely the reason for the performance drop.

My question is, couldn't `manipulate!` just use [MiniMagick::Image.new](http://www.rubydoc.info/github/minimagick/minimagick/MiniMagick%2FImage%3Ainitialize) since the original file is [being overridden anyway](https://github.com/carrierwaveuploader/carrierwave/blob/v0.11.2/lib/carrierwave/processing/mini_magick.rb#L265)?

Also, having different versions in the Uploader already makes copies with prefixes such as `cover_` *before* running `manipulate!`, doesn't it?

There might be other use cases that would be affected by this change, but I'm not sure. If I have more time, I can investigate the implications further.

### LINKED ISSUE #2157: MiniMagick `resize_and_pad` does not pad
```ruby
  ...

  include CarrierWave::MiniMagick

  version :canonized_logo do
    process :resize_and_pad => [800, 400, "#000000", Magick::CenterGravity]
  end

  ...
```

INPUT: 626x416 image

lion-cubs

EXPECTED OUTPUT: 800x400 image

ACTUAL OUTPUT: 602x400 image
output

Makes no sense. No padding occurred at all.

Why is this?

janko:
[ImageProcessing](https://github.com/janko-m/image_processing) gem is a wrapper gem around MiniMagick and libvips, which implements common resizing macros. Instead of the CarrierWave project having to maintain those macros, it can use the ImageProcessing to do the heavy lifting for us. The ImageProcessing gem will constantly get updated with best practices, so any bug fixes and performance improvements there will automatically benefit CarrierWave.

Some improvements that ImageProcessing gem adds on top of MiniMagick is automatic [orientation](https://www.imagemagick.org/script/command-line-options.php#auto-orient) and thumbnail [sharpening](https://www.imagemagick.org/script/command-line-options.php#sharpen). I just published a [blog post](https://twin.github.io/imageprocessing-1-0-released/) about ImageProcessing.

This PR still maintains backwards compatibility by leaving the `#manipulate!` method around, and yielding the `MiniMagick::Image` object in processing methods (even though that should be considered deprecated). But it's recommended to use `#minimagick!`, as it exposes the much more flexible `ImageProcessing::Builder` object.

The ImageProcessing gem automatically pulls in `mini_magick` as a dependency, so this step can now be removed from the user setup.

Should fix #2157
Fixes #2062


0rvar:
It would be great to get this PR fixed up and merged

mshibuya:
Awesome, thanks!

PR #2299, opened 2018-04-06
use cached file first to open image with MiniMagic

### LINKED ISSUE #1947: Problem with MiniMagick height and width methods
When I try to get the width or height of an image (added here: https://github.com/carrierwaveuploader/carrierwave/pull/1405), I receive the following error:

``` Ruby
> image.height
Errno::ENOENT: No such file or directory @ rb_sysopen - /uploads/profile/image/853/Screen_Shot_2016-05-03_at_20.54.29.png
from /usr/local/opt/rbenv/versions/2.3.0/lib/ruby/2.3.0/open-uri.rb:37:in `initialize'
```

which is absolutely right. There is no `/uploads` (note the leading `/`). I think it's because of [this line](https://github.com/carrierwaveuploader/carrierwave/blob/ef0d36a7c195621d8d7d4c769a47a9aee6f64d95/lib/carrierwave/processing/mini_magick.rb#L320), because `.url` returns `/uploads/profile/image/853/Screen_Shot_2016-05-03_at_20.54.29.png`, while something like `.path` or `.current_path` returns the full path: `/full/path/to/uploads/profile/image/853/Screen_Shot_2016-05-03_at_20.54.29.png`.

Does `.url` ever return something falsey, so the method would work?

Using:
Rails 4.2.6
carrierwave `master` version.
MiniMagick
Ruby 2.3.0

pan:
When a file is uploaded and it is still not saved yet, getting its width or height fails with an error message something like in #1947. So this fix uses the cached file first if it is available or use previous strategy to continue.

This pull request fixed #1947.

### DIFF
--- a/lib/carrierwave/processing/mini_magick.rb
+++ b/lib/carrierwave/processing/mini_magick.rb
@@ -341,11 +341,8 @@ module CarrierWave
       end
 
       def mini_magick_image
-        if url
-          ::MiniMagick::Image.open(url)
-        else
-          ::MiniMagick::Image.open(current_path)
-        end
+        img = cached? ? full_cached_path : url || current_path
+        ::MiniMagick::Image.open(img)
       end
 
   end # MiniMagick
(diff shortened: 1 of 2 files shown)


mshibuya:
Could you write the test for this fix?

pan:
I hope I can find some time to fix it in future.

mshibuya:
Closing in favor of #2349.

PR #2303, opened 2018-04-12
Fix #1780 (Fog based file upload streaming)

### LINKED ISSUE #1780: large file is going to blow memory
I believe the commit 927479c84e968ed03546c7fdd16107beeb771a22 (PR #1517 and issue #1338) leads to a memory-problem when uploading large files: 
I guess `file.read` reads the complete file-content into memory. Fog can not "stream" the file from disk, using the multipart-feature correctly (when set `fog_attributes[:multipart_chunk_size]`)
At least, when I revert this commit, my large files (more GB than available mem) are uploaded. Current master returns a "NoMemoryError: failed to allocate memory".

DarkArc:
## What this is about

This fixes an issue with file upload streaming. Previously passing the file was added in PR #468, and was subsequently broken by PR #1517.

This corrects the approach implemented in PR #468, without introducing the problem in PR #1517.

## Approaches

I came up with three different approach for file upload:

### Approach A

We could read the file into memory as carrierwave currently does. This results in large files often exhausting the heap space of the process, thus making carrierwave a poor choice for large file uploads.

### Approach B

We could set a flag to mark a stream uploaded file as needing retrieved. This would result in an additional retrieval requests carrierwave does not currently make, potentially costing in bandwidth for users making use of the current code.

### Approach C

We could store a reference to the source file, and then use that for reading. This would result in issues reading the uploaded file content if the source file is a temp file that goes out of scope.

## Chosen approach
I ended up using Approach B, as I think it's the least buggy/best for most people. However, I'm open to feedback, my first commit in this PR is Approach C, however, the second commit changes it back to Approach B.

I wanted to open up a pull request to see what the general carrierwave community thought was the best approach to fix streaming support.

### DIFF
--- a/lib/carrierwave/storage/fog.rb
+++ b/lib/carrierwave/storage/fog.rb
@@ -276,7 +276,10 @@ module CarrierWave
         #
         # [String] contents of file
         def read
-          file.body
+          # Use the file body if this is a get request
+          # otherwise recycle the reference to the uploaded file.
+          return file.body unless @source_file
+          @source_file.read
         end
 
         ##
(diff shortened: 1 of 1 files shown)


mshibuya:
#2314 have been merged instead.

PR #2304, opened 2018-04-16
[Jin] - fix missing error messages

sunjintw:
fix missing error messages of mime_type_pattern_white_list_error and mime_type_pattern_black_list_error


mshibuya:
MagicMimeBlacklist/MagicMimeWhitelist are old and not to be used, use ContentTypeBlacklist/ContentTypeWhitelist instead.

PR #2306, opened 2018-04-19
More langs

vguti:
es, de, it, fr languages added.


mshibuya:
Thanks for the PR, but we maintain all translations in [carrierwave-i18n](https://github.com/carrierwaveuploader/carrierwave-i18n).

PR #2307, opened 2018-04-23
Update README.md

hcfairbanks:
Adds link to minitest gem.


mshibuya:
Added the link with a slight change of location, thanks!

PR #2310, opened 2018-04-27
change more explicit comment

ogelacinyc:
'name_of_directory' to 'name_of_bucket' about fog AWS S3 configuration

I think that it is a explicit expression
because aws is using the bucket not the directory


mshibuya:
👍

PR #2331, opened 2018-07-29
Check if before is either nil or empty, not just nil in lib/carrierwave/mounter.rb

rbclark:
It doesn't make sense to me to check here if `before` is only nil, the check should be if it is nil or empty, since before being an empty array defeats the purpose of the rest of this method.

This is causing a problem with https://github.com/diogob/carrierwave-postgresql/issues/33 for me, where the second part of this method breaks hard when before is empty.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ gemfile:
 sudo: false
 
 before_install:
-  - gem install bundler
+  - gem install bundler -v 1.16.2
 
 before_script:
   - psql -c 'create database carrierwave_test;' -U postgres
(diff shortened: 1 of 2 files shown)


rbclark:
It looks like the builds for this are passing on everything except ruby 2.4 and 2.5, and they seem to be failing due to some cucumber failure. I don't think my change is the cause of those failures, really not sure what is going on there.

mshibuya:
Looks good, but we need test code which covers this fix.

nononoy:
Any news here?

rbclark:
I'm not exactly sure how I would go about testing this, I spent some time trying to understand the test suite today and I'm not sure I found a good place to integrate it. The issue itself that this was trying to solve doesn't even really affect core carrierwave, it only affects the carrierwave-postgresql uploader.

mshibuya:
Your solution does not help, according to diogob/carrierwave-postgresql#33.

Issue #2343, opened 2018-10-15
Can not assign nil to remote url


adelnabiullin:
Hello.

Here you assign `urls` value to `@remote_urls` only if it exists:
https://github.com/carrierwaveuploader/carrierwave/blob/f32db18a3cd45a55d4e008990a814438a959ed6e/lib/carrierwave/mounter.rb#L73-L75

But why not to `@remote_urls = nil` if condition's value is true?

For example, if downloading or processing fails I want to be able to set remote url's value to nil to save my object totally without the image (using rescue), but I just can not do it.

Thanks.

adelnabiullin:
I've tried to change CarrierWave code: assign @remote_url to nil as I suggested in the issue topic, but it did not help me, I just encountered the following:
Steps:
```
object.remove_image!
object.remote_image_url = nil
object.errors.clear
object.save!
```
Result: saw displayed ActiveRecord validation errors related with CarrierWave download/processing errors as before.

So I am still waiting any response with answer on - How can I do the following thing?

If downloading or processing fails I want to be able to set remote url's value to nil to save my object totally without the image.

mshibuya:
Dup of #2067.

PR #2345, opened 2018-10-16
Clear remote urls on assigning nil to remote_[:column]_url

adelnabiullin:
Add possibility to clear remote urls list on assigning nil to remote_[:column]_url
Issue: https://github.com/carrierwaveuploader/carrierwave/issues/2343

Usecase: If downloading or processing fails I want to be able to set remote url's value to nil to save my object totally without the image (using rescue).


adelnabiullin:
Unfortunately it does not help

PR #2357, opened 2018-12-17
Fix failing master builds

stanhu:
Leftover `new_dir` directories with the wrong permissions were causing
false positives when multiple examples were running.

### DIFF
--- a/spec/sanitized_file_spec.rb
+++ b/spec/sanitized_file_spec.rb
@@ -9,8 +9,11 @@ end
 
 describe CarrierWave::SanitizedFile do
   before do
-    FileUtils.rm_rf(file_path("new_dir"))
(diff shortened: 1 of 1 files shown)


spec/sanitized_file_spec.rb
```diff
@@ -9,6 +9,7 @@
 
 describe CarrierWave::SanitizedFile do
   before do
+    FileUtils.rm_rf(file_path("new_dir"))
```
mshibuya: Hooking into `after` might be better, where cleanup operations are supposed to be.

mshibuya:
Thanks for the fix!

stanhu:
@mshibuya Would you mind tagging a new version? Thanks!

stanhu:
Actually, I was waiting for https://github.com/carrierwaveuploader/carrierwave/pull/2332 too.

Issue #2354, opened 2018-12-14
Incorrect url generated for fog storage + aws


ransombriggs:
We are in the process of moving to carrierwave but have found an issue in url generation & fog when the bucket has a period in it like `foo.bar`.  The problem is that carrierwave detects as a [valid subdomain](https://github.com/carrierwaveuploader/carrierwave/blob/800e0db2554a06993f6026abcec00e7b46b47ceb/lib/carrierwave/storage/fog.rb#L363-L368) where as fog [does not](https://github.com/fog/fog-aws/blob/master/lib/fog/aws/storage.rb#L285-L288).  I tested using the `https://foo.bar.s3.amazonaws.com/` style url that carrierwave constructs but when I request it I get a certificate error because of the multiple nestings of domains causing the cert to not match.

```
curl: (51) SSL: certificate subject name (*.s3.amazonaws.com) does not match target host name 'foo.bar.s3.amazonaws.com'
```

I can file a pull request for this and fix in either one of two ways, either I could add the extra condition like fog does, or I could add a fog option to force the generation to [pass to fog](https://github.com/carrierwaveuploader/carrierwave/blob/800e0db2554a06993f6026abcec00e7b46b47ceb/lib/carrierwave/storage/fog.rb#L375-L376).  Let me know what you think and I can put something together.

ransombriggs:
I wrote a monkey patch in my project to get this working and I think the better way would be to add the extra conditional to avoid the extra network calls, will work on a PR next week.

ransombriggs:
I finished our transition to carrierwave so I went ahead and filed a [PR](https://github.com/carrierwaveuploader/carrierwave/pull/2359) for this bug.

PR #2359, opened 2018-12-20
Fix issue with subdomains and buckets with periods

ransombriggs:
This brings over the functionality from [fog](https://github.com/fog/fog-aws/blob/master/lib/fog/aws/storage.rb#L285-L288) that avoids using the aws subdomain when https and the bucket has periods in it.  I had to update the tests for `assets.site.com` because this also falls victim to the same [bug that I encountered](https://github.com/carrierwaveuploader/carrierwave/issues/2354).

```
-> % curl https://assets.site.com.s3.amazonaws.com
curl: (51) SSL: certificate subject name (*.s3.amazonaws.com) does not match target host name 'assets.site.com.s3.amazonaws.com'
```


mshibuya:
Sounds reasonable, thanks!

PR #2377, opened 2019-02-26
Pass OpenStack options in get_object_https_url

stanhu:
The options parameter has been supported all along in OpenStack. This
may be needed to talk to servers that are configured on non-standard
ports.


mshibuya:
Thanks :+1:

PR #2381, opened 2019-03-19
Add support Aliyun provider to authenticated_url aliyun

Nitrino:
Adds authenticated_url support when using [fog-aliyun](https://github.com/fog/fog-aliyun)

### DIFF
--- a/lib/carrierwave/storage/fog.rb
+++ b/lib/carrierwave/storage/fog.rb
@@ -177,7 +177,7 @@ module CarrierWave
 
         ##
         # Return a temporary authenticated url to a private file, if available
-        # Only supported for AWS, Rackspace, Google and AzureRM providers
+        # Only supported for AWS, Rackspace, Google, AzureRM and Aliyun providers
         #
         # === Returns
         #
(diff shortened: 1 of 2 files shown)


lib/carrierwave/storage/fog.rb
```diff
@@ -204,6 +204,9 @@ def authenticated_url(options = {})
                 connection.get_object_https_url(@uploader.fog_directory, path, expire_at, options)
               when 'OpenStack'
                 connection.get_object_https_url(@uploader.fog_directory, path, expire_at)
+              when 'Aliyun'
+                expire_at = expire_at - Time.now
```
Nitrino: We should calculate `expire_at ` from now because inside aliyun-fog added Time.now
https://github.com/fog/fog-aliyun/blob/6f4d1716488bbcb09ba6538eb3d1cc07d2f26f4b/lib/fog/aliyun/requests/storage/get_object_http_url.rb#L25

mshibuya: Could you add a test for this feature?

Nitrino: What does this logic require testing?
expire_at is defined in this function and passed to another library.
I see here only the opportunity to test that the `url` method with the correct parameter was called inside the `authenticated_url` method. But this test is not of great use.

mshibuya:
fog-aliyun does not support fog mocking, so writing spec is not so easy.
Merging this for now, thanks.

PR #2383, opened 2019-03-21
Preserve connection cache when setting fog_credentials

dmitryshagin:
Create a setup using this wiki https://github.com/carrierwaveuploader/carrierwave/wiki/How-to:-Define-different-storage-configuration-for-each-Uploader leads to initializing CarrierWave on every uploader initialization instead of using cached connections.
This PR adds connection cache utilization for the case described in wiki above.


mshibuya:
Makes sense, thanks!

PR #2386, opened 2019-04-01
CI: Update matrix

olleolleolle:
This PR uses latest JRuby in CI matrix.

  - Also: Removes an old setting from Travis. See https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration


mshibuya:
Thanks ❤️

PR #2401, opened 2019-06-19
Rework multiple file upload

### LINKED ISSUE #2393: Rework multiple file upload
There's many quirkiness in current multiple file upload feature. It should come with

- ability to append a file without discarding existing ones
- ability to delete an individual file
- ability to reorder files without re-uploading

in the next release.

mshibuya:
Closes #2393


mshibuya:
Here we go!

PR #2403, opened 2019-07-03
base64 encoding

ryan-lingle:
problem
---------
- run into CORS issues when attempting to modify an image whose src is from another domain
    - ex: cropping a profile pic for a user while storing images with S3
    - ex error: '...Tainted canvases may not be exported.'

solution
--------
- in my project, I overwrote the to_json module to serialize into base64 in the appropriate image src format
- but in this pr, I created a separate method

notes
------
- the current implementation could certainly use some refactoring (for example it would probably blow up if url was nil) but IMO the general feature is useful
- perhaps an additional key could be added to the JSON hash

### DIFF
--- a/lib/carrierwave/uploader/serialization.rb
+++ b/lib/carrierwave/uploader/serialization.rb
@@ -23,6 +23,17 @@ module CarrierWave
         serializable_hash.to_xml(merged_options)
       end
 
+      def to_base64
+        {"data" => as_base64(url)}.merge Hash[versions.map { |name, version| [name.to_s, { "data" => as_base64(version.url) }] }]
+      end
+
+      private
+
+      def as_base64(url)
+        bytes = open(url) {|f| f.read }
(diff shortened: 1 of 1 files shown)


mshibuya:
Sounds like not the right solution for the problem, did you try this?
https://stackoverflow.com/questions/22710627/tainted-canvases-may-not-be-exported/22713161

ryan-lingle:
Yeah, adding the `crossOrigin="Anonymous"` tag results in a bad request. The S3 server response would need to have a `Access-Control-Allow-Origin "*"` header and they do not. 

Base64 encoding it before json serialization seems to be working fine so far. If it's unwise from a security standpoint I would be grateful to hear why.

mshibuya:
You have to set the header yourself, that should be how this problem is solved.
https://stackoverflow.com/questions/17533888/s3-access-control-allow-origin-header

Issue #2411, opened 2019-08-20
After bundle update, carrier wave stopped uploading images from remote url api 


farooqch11:
NoMethodError: undefined method `[]' for nil:NilClass

mshibuya:
Full stack trace please.

farooqch11:
i downgraded to last version to make it work in production! but this is was not going.

mshibuya:
I need details to track down the issue, what did you do and what happened? How does your code look like and how was the whole stack trace of the error?

farooqch11:
i am using carrier wave with the API , once i use remote_image_url, it throws weird error with latest version! 
- update your bundle
-use api to send remote_url 
-it will break instead of downloading image and uploading again!

mshibuya:
I'm telling you that I need **full stack trace**, please.
That's what you can see like this upon the error, by pressing `Full Trace`.
![](http://cdn-ak.f.st-hatena.com/images/fotolife/y/y_yagi_2/20140810/20140810074654.jpg)

farooqch11:
@mshibuya  i dont have stack trace now, because it was happening on heroku and production! i had to update it quickly without having complaints!

 but i mentioned how to **regenerate** it!

fukayatsu:
In my case, 

    undefined method `[]' for nil:NilClass

occurred at

https://github.com/carrierwaveuploader/carrierwave/blob/v2.0.1/lib/carrierwave/downloader/remote_file.rb#L28 

when the remote image response header includes `content-disposition: inline;filename=""` (from google auth image)

---

```ruby
[1] pry(main)> match = 'content-disposition: inline;filename=""'.match(/filename=(?:"([^"]+)"|([^";]+))/)
=> nil
[2] pry(main)> match[1].presence || match[2].presence
NoMethodError: undefined method `[]' for nil:NilClass
```


----

trace:

```
vendor/bundle/ruby/2.6.0/gems/carrierwave-2.0.1/lib/carrierwave/downloader/remote_file.rb:28:in `filename_from_header': undefined method `[]' for nil:NilClass (NoMethodError)
    from vendor/bundle/ruby/2.6.0/gems/carrierwave-2.0.1/lib/carrierwave/downloader/remote_file.rb:11:in `original_filename'
    from vendor/bundle/ruby/2.6.0/gems/carrierwave-2.0.1/lib/carrierwave/sanitized_file.rb:43:in `original_filename'
[...]

PR #2419, opened 2019-09-10
Fix `remote_x_url=` raise undefined method `[]' for nil:NilClass (NoMethodError)

### LINKED ISSUE #2415: upgrade from 1.3.1 to 2.0.0 breaks remote_x_url when coming from s3
Something weird happens when doing 

```ruby
 order_attachment.remote_document_file_url = template.document_url
 order_attachment.save!
```
where template.document_url is an URL from another uploader (using s3, yes we are basically copying a file from an uploader to an other). We have a `could not download file: 400 Bad Request` error. The weird part is that when using a url from somewhere else (any image on the internet for example) there is no issue.

What could go wrong?

fukayatsu:
Fix `remote_x_url=` when the response header includes `content-disposition: inline;filename=""`

Please see https://github.com/carrierwaveuploader/carrierwave/issues/2411#issuecomment-529735994 for detail.

Maybe related to #2415 also.

### DIFF
--- a/lib/carrierwave/downloader/remote_file.rb
+++ b/lib/carrierwave/downloader/remote_file.rb
@@ -23,10 +23,12 @@ module CarrierWave
       private
 
       def filename_from_header
-        if file.meta.include? 'content-disposition'
-          match = file.meta['content-disposition'].match(/filename=(?:"([^"]+)"|([^";]+))/)
-          match[1].presence || match[2].presence
-        end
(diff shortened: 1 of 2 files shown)


fukayatsu:
I'll rebase this branch after CI fixed :)

mshibuya:
Awesome, thanks!
I'll work on the CI later.

fukayatsu:
Thank you for your hard work!

PR #2438, opened 2019-11-18
Docs: add a complete list of processing methods

jaredbeck:
The docs currently only demonstrate two processing methods, but mini_magick has five.

I don't want to duplicate too much of the "real" documentation (from `carrierwave/processing/mini_magick.rb`) in the readme, but I do want to give people a list of methods, and tell them where to find said real docs.


mshibuya:
Thanks 👍

PR #2439, opened 2019-11-25
Fix S3 path-style URL for host with dots for buckets that are placed in other regions than us-east-1

Bonias:
https://github.com/carrierwaveuploader/carrierwave/pull/2359 added support for bucket names with period, but this only works for `us-east-1` region. For other regions `s3.{aws-region}.amazonaws.com` needs to be used (see https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html#access-bucket-intro).

This PR adds support for bucket names with period that are placed in regions other than `us-east-1`.

---

Also, I'm wondering about one other thing regarding virtual-hosted–style URL. https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html#access-bucket-intro says:

> Buckets created in Regions launched after March 20, 2019 are not reachable via the https://bucket.s3.amazonaws.com naming scheme. 

Should carrierwave be updated to use `http://{bucket}.s3.{aws-region}.amazonaws.com`? If so should it:

  - use `http://{bucket}.s3.{aws-region}.amazonaws.com` for all regions (this is what both `aws-sdk-s3`, `carrierwave-aws` do and kind of what `fog-aws` does; `fog-aws` uses `http://{bucket}.s3-{aws-region}.amazonaws.com`),
  - use `http://{bucket}.s3.{aws-region}.amazonaws.com` only for regions created after `March 20, 2019`? (in that case carrierwave would require to have a list of regions created before `March 20, 2019`)

?

### DIFF
--- a/lib/carrierwave/storage/fog.rb
+++ b/lib/carrierwave/storage/fog.rb
@@ -161,6 +161,8 @@ module CarrierWave
       end
 
       class File
+        DEFAULT_S3_REGION = 'us-east-1'
+
         include CarrierWave::Utilities::Uri
 
         ##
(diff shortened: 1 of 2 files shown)


mshibuya:
Awesome, thanks!

PR #2442, opened 2019-12-18
Provide public interface alternatives to whitelist/blacklist terminology

grantbdev:
The goal of this pull request is to give developers the option to use something like allowlist/blocklist terminology with this gem instead of whitelist/blacklist. https://github.com/rubocop-hq/rubocop/pull/6464 has links to the original motivation and many changes to other gems that have been replacing those words.

Since managing allowed content types and extensions is a significant part of the gem's purpose and impacts the code users write, as a starting point the changes in this PR are intended to be minimally intrusive. Overriding the existing methods still works so it is backwards compatible. If there is buy-in from the maintainers, I will gladly contribute deeper changes to the internals and deprecation of the old methods could be considered as well.

Allowlist and blocklist are only starting points here and I'm open to different names that could be more descriptive. The wording in some places could probably be improved if there is consensus on the alternative. For example, I used "allowlisted" for consistency but you may feel as I do that it sounds awkward and it could be clearer as "allowed" or "permitted" in those instances.

In this case, I would consider naming the methods like `allowed_extensions` and `prohibited_content_types` if you want to match the wording in the [translation file](https://github.com/grantbdev/carrierwave/blob/6545b6bb7e01a53dba465931020ffeb1a46bc885/lib/carrierwave/locale/en.yml#L7-L9) and use language closer to what application users might see.

### DIFF
--- a/lib/carrierwave/uploader/content_type_blacklist.rb
+++ b/lib/carrierwave/uploader/content_type_blacklist.rb
@@ -28,7 +28,11 @@ module CarrierWave
       #       [/(text|application)\/json/]
       #     end
       #
-      def content_type_blacklist; end
+      def content_type_blacklist
+        content_type_blocklist
+      end
+
+      def content_type_blocklist; end
 
     private
 
--- a/lib/carrierwave/uploader/content_type_whitelist.rb
+++ b/lib/carrierwave/uploader/content_type_whitelist.rb
@@ -7,6 +7,8 @@ module CarrierWave
         before :cache, :check_content_type_whitelist!
       end
 
+      def content_type_allowlist; end
+
       ##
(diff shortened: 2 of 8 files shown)


mshibuya:
Thank you for the proposal! I totally agree to merge this change. 
In addition to that, I want to make some changes after merging:

- The pair of allowlist/denylist sounds more natural than allowlist/blocklist. (ref. https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=49decddd39e5f6132ccd7d9fdc3d7c470b0061bb)
- I want whitelist/blacklist to be marked as deprecated so the users can migrate smoothly.

PR #2457, opened 2020-03-14
Fix overly aggressive remote image url escaping in specific instances

### LINKED ISSUE #2456: remote download improperly escapes query params in specific cases
When the remote download is trying to parse and escape the remote absolute image url, it can improperly escape some characters needed in order to work - specifically when the image src is actually in the query string.

Example:

`object.remote_image_url = "https://i.vimeocdn.com/filter/overlay?src0=https%3A%2F%2Fi.vimeocdn.com%2Fvideo%2F821706979_1280x720.jpg&src1=https%3A%2F%2Ff.vimeocdn.com%2Fimages_v6%2Fshare%2Fplay_icon_overlay.png"`

That will actually escape the `https%3A%2F%2F` as `https%253A%252F%252F` and it won't actually download it correctly. The `process_uri` method in the downloader base class needs to take this into account. Right now it escapes all query param `%'`s.

courtsimas:
Closes #2456 - which is an issue where in specific instances, when the remote download is trying to parse and escape the remote absolute image url, it can improperly escape some characters that are actually needed in order for the image/download to work - specifically when the image source is actually in the query string, and not in the base portion of the path.

You'll see this type of pattern semi-often; where the image path/source is actually defined in the query parameter in on-the-fly overlays/mashups of images (which imgix supports) and in the way Vimeo's thumbnail images work as well (among others).

The related issue at #2456 has more information and an example.


lib/carrierwave/downloader/base.rb
```diff
@@ -40,7 +40,9 @@ def download(url, remote_headers = {})
       def process_uri(uri)
         uri_parts = uri.split('?')
         encoded_uri = Addressable::URI.parse(uri_parts.shift).normalize.to_s
-        encoded_uri << '?' << Addressable::URI.encode(uri_parts.join('?')).gsub('%5B', '[').gsub('%5D', ']') if uri_parts.any?
+        encoded_uri << '?' << Addressable::URI.encode(uri_parts.join('?')).
+          gsub('%5B', '[').gsub('%253A%252F%252F', '%3A%2F%2F').
```
courtsimas: This is specifically targeting for the `://` portion of the `https://` or `http://` or `some-protocol://` and removing the double escape.

coneybeare:
Also related to https://github.com/carrierwaveuploader/carrierwave/pull/2473, but different

courtsimas:
Ping. Anyone using Vimeo or imgix for mashups (overlays) defining the secondary image in a query param can’t use carrierwave in this manner.

courtsimas:
@mshibuya since your commit doesn't actually seem to address the issue my PR addressed (albeit, maybe mine wasn't the best approach, but it worked), what's an easy way of overriding the `process_uri` method?

mshibuya:
What do you mean?
Your test case in this PR was already merged in c3b4cc6d00df622fac1d275f62740334e8036ad1, and my commit 3faf7491e33bd10ae8b3e0010501fc96a76c21c3 made it pass.

courtsimas:
Dude I apologize. You’re right. Thanks man.

PR #2460, opened 2020-03-17
fix: default sharpen option in image_processing

StephaneRob:
Prevent image_processing to automatically sharpen images https://github.com/janko/image_processing/pull/24.

To be able to keep control on imagemagick options and add sharpen in `combine_options`, force image_processing `sharpen` option to `false`.


janko:
ImageProcessing 1.11.0 doesn't sharpen by default anymore. CarrierWave can either require that version from now on, or rely on people upgrading to latest ImageProcessing version.

PR #2462, opened 2020-03-26
[1.x-stable] - Fix deprecation warning ruby 2.7

aubinlrx:
This PR fix some deprecation warnings introduced with Ruby 2.7.

- Separation of positional and keyword arguments in Ruby 3.0 [source](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/)
- Deprecation of URI.decode [link](https://docs.knapsackpro.com/2020/uri-escape-is-obsolete-percent-encoding-your-query-string)

```
warning: URI.escape is obsolete
warning: URI.encode is obsolete
```

- Deprecation of Kernel.open

```
calling URI.open via Kernel#open is deprecated, call URI.open directly or use URI#open

### DIFF
--- a/lib/carrierwave/sanitized_file.rb
+++ b/lib/carrierwave/sanitized_file.rb
@@ -312,7 +312,7 @@ module CarrierWave
     def mkdir!(path, directory_permissions)
       options = {}
       options[:mode] = directory_permissions if directory_permissions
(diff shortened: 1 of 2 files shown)


mshibuya:
Thanks!

PR #2464, opened 2020-03-31
Skip content type filters where possible.

alexpooley:
Prior to this patch every content_type_* filter would parse magic headers via `SanitizedFile#content_type`,
even when the filter was not defined.

I have split the primary conditional check in to a guard clause that
simply returns unless the blacklist or whitelist returns truthy.

I've made the same changes to the extension filters for consistency.

I wanted to remove any unnecessary parse of magic headers completely on upload. This is a half measure to getting there.


mshibuya:
Thanks!

PR #2474, opened 2020-04-23
Do not return invalid content type on text files

### LINKED ISSUE #2424: Bug with content type detection
Hi there!

I think it should be nil here : https://github.com/carrierwaveuploader/carrierwave/blob/d12a289d9cdd68d26d3900cf7f1e4e879678d16b/lib/carrierwave/sanitized_file.rb#L334
(like in the rescue block) to let `mini_mime_content_type` have a chance to find the right mime type.

inkstak:
Fixes #2424

It also supports this edge case:
I have to deal with text files provided by public administrations with inconsistent extensions ( .N000777, .BATI, .txt, missing extensions) alongside known extensions (.csv, .pdf, images, etc.)

Returning a missing content type is quite different from returning an invalid content type.

### DIFF
--- a/lib/carrierwave/sanitized_file.rb
+++ b/lib/carrierwave/sanitized_file.rb
@@ -330,9 +330,16 @@ module CarrierWave
 
     def mime_magic_content_type
       if path
-        File.open(path) do |file|
-          MimeMagic.by_magic(file).try(:type) || 'invalid/invalid'
+        type = File.open(path) do |file|
+          MimeMagic.by_magic(file).try(:type)
         end
+
+        if type.nil?
+          type = ::MiniMime.lookup_by_filename(path).try(:content_type)
+          type = 'invalid/invalid' unless type.nil? || type.start_with?('text/')
+        end
+
+        type
       end
     rescue Errno::ENOENT
       nil
--- a/spec/sanitized_file_spec.rb
+++ b/spec/sanitized_file_spec.rb
@@ -221,6 +221,26 @@ describe CarrierWave::SanitizedFile do
       expect(sanitized_file.content_type).to eq 'invalid/invalid'
     end
 
+    it "returns valid content type on text file" do
+      file = File.open(file_path('bork.txt'))
+
+      sanitized_file = CarrierWave::SanitizedFile.new(file)
+
+      expect { sanitized_file.content_type }.not_to raise_error
+
+      expect(sanitized_file.content_type).to eq 'text/plain'
+    end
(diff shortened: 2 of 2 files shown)


n-rodriguez:
Hi there! Any news?

n-rodriguez:
Hi there! Any news?

lib/carrierwave/sanitized_file.rb
```diff
@@ -330,9 +330,16 @@ def existing_content_type
...
       if path
-        File.open(path) do |file|
-          MimeMagic.by_magic(file).try(:type) || 'invalid/invalid'
+        type = File.open(path) do |file|
+          MimeMagic.by_magic(file).try(:type)
         end
+
+        if type.nil?
+          type = ::MiniMime.lookup_by_filename(path).try(:content_type)
+          type = 'invalid/invalid' unless type.nil? || type.start_with?('text/')
```
krasnoukhov, line 339: I'm curious why you didn't go with your [original approach](https://github.com/carrierwaveuploader/carrierwave/issues/2424#issuecomment-618238080) for `image/*` check only? There are a lot of other options like `application/*` or `audio/*` for which the issue will still be relevant

inkstak: The actual strategy on carrierwave is giving high confidence in the result of `MimeMagic.by_magic`:  
* If it returns something : it's valid
* If it returns nothing : it's invalid

The approach I suggested in [#2424](https://github.com/carrierwaveuploader/carrierwave/issues/2424#issuecomment-618238080) doesn't trust the result and still considers missing types as invalid:  
* If it returns an image : we double-check the type from the filename
* If it's present but anything other than an image : it's valid
* If it returns nothing : it's still invalid

The final solution tries to keep confidence in `MimeMagic.by_magic` but considers the nothing case:
* If it returns something : it's still valid.
* If it returns nothing : we double-check the type from the file name:
  * if it's still missing : it's just something unknown so let's return `nil`
  * if it looks like text : the types don't match but there is no risk in considering it valid
  * if it's anything else : it looks like type spoofing so let's return invalid type


I guess images, audio or video files not detected as such by `MimeMagic.by_magic` should be considered at risk as they can then be processed by sub-programs, added and played from HTML pages.

I've mixed feelings about `application/*` because it covers a lot of cases.

zjeraar:
Any chance of merging this?

mshibuya:
Thank you and sorry for delay.

Issue #2488, opened 2020-06-03
content_type_whitelist blocking image/svg+xml


sylvainbx:
I have an uploader with the following mime-type whitelist:
```
  def content_type_whitelist
    %w[image/svg+xml]
  end
``` 
When I upload an svg file, it fails with the following (crazy) error message:
```
{"project_cads.attachment":[
  "You are not allowed to upload image/svg+xml files, allowed types: image/svg+xml"
]}
```

Any idea on how to solve this?

RigoTheDev:
Same issue. Seems like this line is rejecting `image/svg+xml` due to Regex https://github.com/carrierwaveuploader/carrierwave/blob/6c7b126bb4a2bc736d79467dc4a58fa559ac5696/lib/carrierwave/uploader/content_type_whitelist.rb#L43

mshibuya:
Merged #2489.

PR #2489, opened 2020-06-03
handle mimetype validation with symbol + (#2488)

sylvainbx:
A lot a mimetypes match the following scheme: type/subtype+other.
Carrierwave must allow the validation of these types without error.
This patch is providing the support for theses types, along with the tests to validate it


mshibuya:
Awesome, thanks!

mhui:
We prob want this on the denylist as well

PR #2500, opened 2020-07-20
Add CarrierWave::Vips and related specs

rhymes:
This PR adds support for [libvips](https://github.com/libvips/libvips) through [ImageProcessing::Vips](https://github.com/janko/image_processing) and [ruby-vips](https://github.com/libvips/ruby-vips).

A few notes:

- I had to unlock `rmagick` as it's currently at version 4 and it wouldn't build version 2 on my machine
- I added [pry-byebug](https://github.com/deivid-rodriguez/pry-byebug) to development dependencies which makes stepping through the code infinitely easier via `binding.pry`
- libvips differs from ImageMagick in that [resize_and_pad](https://github.com/janko/image_processing/blob/master/doc/vips.md#resize_and_pad) adds a black background by default, not a white one. The background is transparent if the alpha channel is set and supported.

### DIFF
--- a/lib/carrierwave/processing/vips.rb
+++ b/lib/carrierwave/processing/vips.rb
@@ -1,0 +1,330 @@
+module CarrierWave
+
+  ##
+  # This module simplifies manipulation with vips by providing a set
+  # of convenient helper methods. If you want to use them, you'll need to
+  # require this file:
+  #
+  #     require 'carrierwave/processing/vips'
+  #
+  # And then include it in your uploader:
+  #
+  #     class MyUploader < CarrierWave::Uploader::Base
+  #       include CarrierWave::Vips
+  #     end
+  #
+  # You can now use the provided helpers:
+  #
+  #     class MyUploader < CarrierWave::Uploader::Base
+  #       include CarrierWave::Vips
+  #
+  #       process :resize_to_fit => [200, 200]
+  #     end
+  #
+  # Or create your own helpers with the powerful vips! method, which
+  # yields an ImageProcessing::Builder object. Check out the ImageProcessing
+  # docs at http://github.com/janko-m/image_processing and the list of all
+  # available Vips options at
+  # https://libvips.github.io/libvips/API/current/using-cli.html for more info.
+  #
+  #     class MyUploader < CarrierWave::Uploader::Base
+  #       include CarrierWave::Vips
+  #
+  #       process :radial_blur => 10
+  #
+  #       def radial_blur(amount)
+  #         vips! do |builder|
+  #           builder.radial_blur(amount)
+  #           builder = yield(builder) if block_given?
+  #           builder
+  #         end
+  #       end
+  #     end
+  #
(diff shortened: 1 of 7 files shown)


mshibuya:
@rhymes  Thank you for providing a great feature!
I'm willing to merge this, but could you make this PR contain only the commit https://github.com/carrierwaveuploader/carrierwave/pull/2500/commits/4575ffd7e6b0f008b0514f34a4798bc2e0366861 ?
We have moved to GitHub Actions recently, so Travis-related code is no longer used.

rhymes:
Hi @mshibuya, sure! I can do that. Thank you for taking the time to review this

lib/carrierwave/processing/vips.rb
```diff
@@ -0,0 +1,330 @@
...
+    # [CarrierWave::ProcessingError] if manipulation failed.
+    #
+    def manipulate!
+      cache_stored_file! if !cached?
+      image = ::Vips::Image.open(current_path)
+
+      image = yield(image)
+      FileUtils.mv image.path, current_path
+
+      image.run_command("identify", current_path)
```
mshibuya, line 264: Does this really work?
Seems like to be a leftover from CarrierWave::Processing::MiniMagick.

spec/processing/vips_spec.rb
```diff
@@ -0,0 +1,261 @@
...
+      instance.resize_and_pad(200, 200, nil, 'centre', true)
+
+      expect(color_of_pixel(instance.current_path, 0, 0)).to include('none')
+    end
+
+    it "doesn't pad with transparent" do
+      binding.pry
+
+      instance.resize_and_pad(200, 200)
+      instance.convert('png')
```
mshibuya, line 112: I could not make this spec pass(apart from `bindings.pry`), it always breaks with `VipsForeignLoad: images do not match` here.
Do you have any idea?

mshibuya:
I've updated GitHub Actions to use Ubuntu 20.04, so `image_processing/vips requires libvips 8.6+` errors will be gone.

spec/processing/vips_spec.rb
rhymes: I'm going to investigate a bit, as it's been a while since I touched this code :)

lib/carrierwave/processing/vips.rb
rhymes: IIRC I did left it there because my initial goal was to replicate the same interface as the `MiniMagick` module. I added a note in the module's documentation:

>     # NOTE: This method exists mostly for backwards compatibility, you should
>     # probably use #vips!.

rhymes:
@mshibuya merged master in and removed `binding.pry`

mshibuya:
OK, I can take care of the rest.

mshibuya:
Thank you for the contribution! 🎉

rhymes:
Thank you @mshibuya!!

PR #2501, opened 2020-07-21
Add rmagick's original error to I18n message

arthurmde:
The file [lib/carrierwave/processing/rmagick.rb](lib/carrierwave/processing/rmagick.rb) adds the RMagick's original error to the I18n message. It is very useful for understanding low-level issues on processing images with image magick, such as [policy errors](https://stackoverflow.com/questions/52998331/imagemagick-security-policy-pdf-blocking-conversion)


mshibuya:
In the past we have removed error message because of security concern: https://github.com/carrierwaveuploader/carrierwave/commit/96314bb42c32d6b24278474e1c877b53a88bfaf8
It needs to be cared to do so.

PR #2503, opened 2020-07-24
CarrierWave::Storage::Fog#copy_to with Content-Type header

### LINKED ISSUE #2484: Uploaded files always have content type "application/json"
I'm using Carrierwave with the fog openstack storage driver (swift).

Version: 2.1.0

Config:

```ruby
  config.storage = :fog

  config.fog_credentials = {
      :provider            => 'OpenStack',
      :openstack_username  => ENV.fetch("CARRIERWAVE_OPENSTACK_USERNAME"),
      :openstack_api_key   => ENV.fetch("CARRIERWAVE_OPENSTACK_API_KEY"),
      :openstack_auth_url  => 'https://auth.cloud.ovh.net/v3/',
      :openstack_region    => 'SBG'
  }

  config.fog_directory = ENV.fetch("CARRIERWAVE_FOG_DIRECTORY")
  config.fog_attributes = {
    'Cache-Control'=>'max-age=315576000', 
    'X-Detect-Content-Type' => 'true'
  }

  config.cache_storage = :fog
  config.fog_public = true
end

class CarrierWave::Storage::Fog::File
  # undefined method `original_filename' for #<CarrierWave::Storage::Fog::File:0x00007fe0e6d6bae0>
  def original_filename
    ::File.basename(path)
  end

  # undefined method `basename' for #<CarrierWave::Storage::Fog::File:0x00007fe0e6d6bae0>
  def basename
    ::File.basename(path)
  end
end
```

The monkey patch for `basename` and `original_filename` seem necessary, because somewhere in the library code it get used...

Files are getting uploaded, and deleted correctly but new files are always content type "application/json".

We have a pretty complicated nested version setup, so its a bit hard to showcase here. But the essence is, that calling `CarrierWave::Storage::Fog::File#content_type` returns nil while `recreate_versions!` or when uploading new videos.

I also tried to manually set the content type via `CarrierWave::Storage::Fog::File#content_type=` with a `process` callback. That has no effect either.

The best option for me would be to always set the `X-Detect-Content-Type` header so that i don't have to deal with this. But the configuration does not apply the header to my blobs,

Described [here](https://docs.openstack.org/api-ref/object-store/index.html?expanded=create-or-replace-object-detail)

Best,
Niklas

### LINKED ISSUE #2487: X-Copy-From request sends "application/json" as content type
I am processing videos with carrierwave and fog as storage provider. All videos that are uploaded have the content type "application/json". I already got so far, that i have overriden the `CarrierWave::Storage::Fog::File#content_type` method but with no success. With that tried i've come to the conclusion, that there is something wrong underneath my processing code.

What i can see with `EXCON_DEBUG=true` is, that the first request, putting my video into a temp file, works as expected:

```
:method              => "PUT"
:headers             => {
    "Accept"         => "application/json"
    "Content-Length" => 3267624
    "Content-Type"   => "video/quicktime"
    "Host"           => "storage.sbg.cloud.ovh.net:443"
    "User-Agent"     => "fog-core/2.2.0"
    "X-Auth-Token"   => "XXXXXXXX"
}
```

I get a 201 Created back. After that, carrierwave processes the file and moves it with fog-openstack into is real destination:

```
:headers             => {
    "Accept"       => "application/json"
    "Content-Type" => "application/json"
    "Host"         => "storage.sbg.cloud.ovh.net:443"
    "User-Agent"   => "fog-core/2.2.0"
    "X-Auth-Token" => "XXXXXXXX"
    "X-Copy-From"  => "/XXXX_development/uploads/tmp/1590760935-202723548356984-0001-0409/loading_screen_flickers.mov"
  }
```

The response for the copy request looks as following:

```
  :headers       => {
    "Content-Length"              => "0"
    "Content-Type"                => "text/html; charset=UTF-8"
    "Date"                        => "Fri, 29 May 2020 14:02:20 GMT"
    "Etag"                        => "e1340ff8ddfb1a8ae1252bc55dae7f85"
    "Last-Modified"               => "Fri, 29 May 2020 14:02:20 GMT"
    "X-Copied-From"               => "XXX_development/uploads/tmp/1590760935-202723548356984-0001-0409/loading_screen_flickers.mov"
    "X-Copied-From-Account"       => "XXXXXXXXXXXX"
    "X-Copied-From-Last-Modified" => "Fri, 29 May 2020 14:02:18 GMT"
    "X-IPLB-Instance"             => "12631"
    "X-Openstack-Request-Id"      => "tx3905d22b0674473e9bd0f-005ed115eb"
    "X-Trans-Id"                  => "tx3905d73b0671173e9bd0f-005ed115eb"
  }
```

After that, the file which got moved will only be accessible via content type "application/json".

Any ideas?

Best,
Niklas

ZuevEvgenii:
Set right Content-Type when copy files from tmp/ directory. 

Resolve #2487
Resolve #2484 
Resolve #2504


lib/carrierwave/storage/fog.rb
```diff
@@ -450,7 +450,10 @@ def filename(options = {})
         # @return [CarrierWave::Storage::Fog::File] the location where the file will be stored.
         #
         def copy_to(new_path)
-          connection.copy_object(@uploader.fog_directory, file.key, @uploader.fog_directory, new_path, acl_header)
+          options = {}
+          options.merge!(acl_header) if acl_header.present?
+          options['Content-Type'] ||= content_type if content_type
```
stanhu, line 455: I also saw that we're missing `@uploader.fog_attributes`. I think we should break this into a separate method? (https://github.com/carrierwaveuploader/carrierwave/pull/2504)

ZuevEvgenii: Thank you, merged

begotten63:
We are looking forward to this fix. Any thoughts when this can be merged into the master?

mshibuya:
Merged in, thanks!

PR #2510, opened 2020-09-24
Fix return value of `cached?` according to document

kmiyake:
The document says `cached?` is expected to return `Bool` value. But actually, the method returns a `@cache_id`.
According to the document and ruby basic style, fixes return value to `Bool` value from `@cache_id`.

This change has breaking changes, so perhaps should not be changed this behavior.


mshibuya:
Makes sense, thanks!

PR #2519, opened 2020-11-13
Performance improvement, fixes twice #url method call

skyeagle:
Hi there,

Currently, the method #url is being called twice (with and without options) in a case of arity != 0 (very common case).
It is very inefficient for signed URLs specifically.

This PR fixes that.

Thanks!


mshibuya:
Thanks!

Issue #2426, opened 2019-09-26
Migration from 1.3.1 to 2.0.1 breaks public permissions with Google Cloud Storage


nerf-qh:
After migration from 1.3.1 to 2.0.1 public permissions per object get 
`ruby 2.6.4`
`carrierwave 2.0.1`
`fog-core 2.1.0`
`fog-google 1.9.1`

example
https://github.com/nerf-qh/carrierwave_update

Config
```
creds = {
  provider: 'Google',
  google_project: ENV['GCS_PROJECT'],
  google_json_key_string: ENV['GCS_AUTH']
}

CarrierWave.configure do |config|
  config.fog_credentials = creds
  config.fog_directory  = ENV['GCS_BUCKET']
  config.fog_public     = true
  config.storage = :fog
end
```

The initial file is public, but in the `copy_to` method this option is lost
https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/storage/fog.rb#L444-L446

The same happens if you just copy files
https://gist.github.com/nerf-qh/2606513f42e3be169e52cac4ca72f5ce

yosiat:
Getting the same issue, when upgrading carrierwave.

We noticed the difference is that:
- old carrierwave gave this ACL:
```
  {
    "entity": "allUsers",
    "role": "READER"
  }
```
- new carrierwave don't give this ACL and this is the only reader role it gives (replaced values):
```
  {
    "entity": "XXXXX",
    "projectTeam": {
      "projectNumber": "XXXXX",
      "team": "viewers"
    },
    "role": "READER"
  },
```

This is a major issue when upgrading carrierwave.

yosiat:
Did investigation of the issue and as written here it happens because of `copy_object` (which I think happens because of - https://github.com/carrierwaveuploader/carrierwave/issues/2136)

I opened an issue to ask fog-google on how to pass auth query parameters so we can pass it from carrierwave - https://github.com/fog/fog-google/issues/505

PR #2525, opened 2020-12-30
GCS: pass acl_header for public read if fog is public

yosiat:
When upgrading from carrierwave 1.3 to 2 we noticed that when we upload to GCS the URL we get is not public, even though we have `fog_public = true` in our initializer.

After some investigation it looks like the copy_object is not maintaining the ACL, this change is passing the correct ACL if `fog_public` is true.

Tested locally and it fixed the issue.

Closes https://github.com/carrierwaveuploader/carrierwave/issues/2426
fog-google pr: https://github.com/fog/fog-google/pull/513

### DIFF
--- a/lib/carrierwave/storage/fog.rb
+++ b/lib/carrierwave/storage/fog.rb
@@ -497,6 +497,8 @@ module CarrierWave
         def acl_header
           if fog_provider == 'AWS'
             { 'x-amz-acl' => @uploader.fog_public ? 'public-read' : 'private' }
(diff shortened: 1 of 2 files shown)


yosiat:
@mshibuya hi! can you look at this PR please?

thiagofm:
Thanks for your contribution. LGTM

yosiat:
@thiagofm thanks a lot :) 

once we have a release, we will upgrade asap :)

PR #2526, opened 2021-01-06
Enable multipart and multithreaded uploads in AWS

stanhu:
When files are stored in object storage, CarrierWave will copy these
files from their temporary location to a final location. However, if the
file is above 5 GB, the upload will fail outright because AWS requires
multipart uploads to be used to copy files above that limit.

Even if multipart uploads were used, files containing several gigabytes
of data may take a long time. In one test, a 6 GB file took several
minutes to copy with fog-aws, while it only took 36 seconds with the aws
CLI. The main difference: multithreading.

fog-aws now supports multipart, multithreaded uploads per these pull
requests:

* https://github.com/fog/fog-aws/pull/578
* https://github.com/fog/fog-aws/pull/579

For this to work, we also need to patch CarrierWave to use the
`File#copy` method instead of the Fog connection `copy_object`
method. We use a concurrency of 10 threads because this is what the AWS
SDK uses, and it appears to give good performance for large uploads.

### DIFF
--- a/lib/carrierwave/storage/fog.rb
+++ b/lib/carrierwave/storage/fog.rb
@@ -450,7 +450,20 @@ module CarrierWave
         # @return [CarrierWave::Storage::Fog::File] the location where the file will be stored.
         #
         def copy_to(new_path)
-          connection.copy_object(@uploader.fog_directory, file.key, @uploader.fog_directory, new_path, acl_header)
+          # fog-aws v3.7.0 supports multithreaded, multipart uploads
+          if fog_provider == 'AWS' && file.respond_to?(:concurrency=)
+            # AWS SDK uses 10 threads by default and a multipart chunk size of 10 MB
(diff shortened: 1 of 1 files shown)


thiagofm:
Hmm. Could you add tests and perhaps have that fog patch in a different class? It's a bit strangey patch. Couldn't this be somehow in fog?

stanhu:
Apparently not all S3 providers (e.g. old Ceph backends) support the Multipart Upload Part API.

mshibuya:
I don't think this to be something that CarrierWave should have. If you believe that using the multithreaded copy is **always** the fastest option, you should submit a PR to fog-aws to change so.
(And I don't think so, I suppose that in typical situation where source and destination objects reside in the same bucket `#copy_object` is faster because it can be processed within AWS S3 and does not need the data to be downloaded and uploaded.)

PR #2532, opened 2021-02-21
Use Fog::Storage::GoogleJSON::Files#each_file_this_page

tamatsu:
I tried to clean my cache files on Google Cloud Storage, but I got this error.
I think I can solve this error by using `each_file_this_page` instead of `each`.
(Fog::Storage::GoogleJSON::Files#each may ignore `:prefix` specification) 

・storage :fog

```
irb(main):002:0> MyUploader.storage.new(MyUploader.new).clean_cache! 0
Sending HTTP get https://storage.googleapis.com/storage/v1/b/...
200
...
Traceback (most recent call last):
        2: from (irb):2
        1: from (irb):2:in `rescue in irb_binding'
NoMethodError (undefined method `map' for nil:NilClass)
```

Thanks.

### DIFF
--- a/lib/carrierwave/storage/fog.rb
+++ b/lib/carrierwave/storage/fog.rb
@@ -145,7 +145,7 @@ module CarrierWave
         connection.directories.new(
           :key    => uploader.fog_directory,
           :public => uploader.fog_public
-        ).files.all(:prefix => uploader.cache_dir).each do |file|
+        ).files.all(:prefix => uploader.cache_dir).each_file_this_page do |file|
           # generate_cache_id returns key formated TIMEINT-PID(-COUNTER)-RND
           time = file.key.scan(/(\d+)-\d+-\d+(?:-\d+)?/).first.map { |t| t.to_i }
           time = Time.at(*time)


mshibuya:
I don't understand what this change fixes, could you provide a test code for this?

lib/carrierwave/storage/fog.rb
```diff
@@ -145,7 +145,7 @@ def clean_cache!(seconds)
         connection.directories.new(
           :key    => uploader.fog_directory,
           :public => uploader.fog_public
-        ).files.all(:prefix => uploader.cache_dir).each do |file|
+        ).files.all(:prefix => uploader.cache_dir).each_file_this_page do |file|
```
mshibuya, line 148: Use of `each_file_this_page` disables pagination handling for [fog-aws](https://github.com/fog/fog-aws/blob/master/lib/fog/aws/models/storage/files.rb#L42-L57).
This code is shared across multiple fog providers, you need to make sure that all of them does not break.

tamatsu:
Thank you for your review.

Here is the reproduction code for Rails.
(I use `gem "carrierwave", '~> 2.1.0'`, because I failed to store files to GCS with `2.2.0`. I'm thinking to create another issue.)

```
$ rails -v
Rails 6.1.3
$ rails c
> uploader = MyUploader.new
> file = Rack::Test::UploadedFile.new(File.join('plain.txt'), 'text/plain')
> uploader.store!(file)
> uploader.cache!(file)
> MyUploader.clean_cached_files! 0
...
Traceback (most recent call last):
        1: from (irb):6
NoMethodError (undefined method `map' for nil:NilClass)
```

app/uploaders/my_uploader.rb
```
  storage :fog

  def store_dir
    "uploads"
  end
end
```

config/initializers/carrierwave.rb
```
  config.fog_provider = 'fog/google'
  config.fog_credentials = {
      provider:               'Google',
      google_project:         'xxx',
      google_json_key_location: 'cred.json'
  }
  config.fog_directory = 'xxx-bucket'
end
```

Gemfile
```
gem "carrierwave", '~> 2.1.0'
gem "fog-google"
```

### Patch
This patch solved the problem.
```
module CarrierWave
  module Storage
    class Fog < Abstract
      def clean_cache!(seconds)
        connection.directories.new(
          :key    => uploader.fog_directory,
          :public => uploader.fog_public
        ).files.all(:prefix => uploader.cache_dir).each_file_this_page do |file|
          # generate_cache_id returns key formated TIMEINT-PID(-COUNTER)-RND
          return unless file.key.match(/\d+-\d+-\d+(?:-\d+)?/)

          time = file.key.scan(/(\d+)-\d+-\d+(?:-\d+)?/).first.map { |t| t.to_i }
          time = Time.at(*time)
          file.destroy if time < (Time.now.utc - seconds)
        end
      end
    end # Fog
  end # Storage
end # CarrierWave
```

### Test code
Writing a test code to delete GCS files seems difficult for me. sorry.

Thank you.

lib/carrierwave/storage/fog.rb
tamatsu: I couldn't understand how CarrierWave::Storage::Fog#clean_cache! uses pagination handling, sorry.
It would be helpful if you could explain in a little more detail.
Thanks.

mshibuya:
You should have something not generated as CarrierWave cache in your cache_dir, that's causing your problem.
I've fixed the issue in 42c620a, thanks for reporting anyway.

PR #2538, opened 2021-03-01
Add "public" messages to errors

akihikodaki:
This pull request supersedes #2499.

The message for `CarrierWave::ProcessingError` used to include
- the name of the underlying image processor (RMagick or MiniMagick) and
- the original error message produced by it.

The message is used for ActiveRecord validation, and it is often shown for an end-user. This behavior has some downsides:
- The message may look cryptic and is confusing for end-users. A typical end-user has no idea what is RMagick, for example.
- The information the message carries may be valuable for an adversary. Especially, the image processor may not be careful enough to mask sensitive information in an error message it produces. e.g. MiniMagick may include stderr of ImageMagick, which may contain file paths and reveal underlying ImageMagick configurations, in an error message.

This change solves the problems by adding `public_message` attributes, which does not contain such sensitive information, to error types and exposing them in ActiveRecord validation.

### DIFF
--- a/lib/carrierwave/error.rb
+++ b/lib/carrierwave/error.rb
@@ -1,8 +1,76 @@
 module CarrierWave
   class UploadError < StandardError; end
-  class IntegrityError < UploadError; end
+
+  class IntegrityError < UploadError
+    # @return [String] Returns the error message which can safely be exposed to
+    # the user.
+    attr_reader :public_message
+
+    # Construct a new IntegrityError object, optionally passing in messages.
+    #
+    # @param error_message [String] Optional message.
+    #
+    # @param public_error_message [String] Error message which can safely be
+    # exposed to the user. If it is not specified, it fallbacks to
+    # error_message. If error_message is not specified, it fallbacks to
+    # a default message translated for the current locale.
+    def initialize(error_message = nil, public_error_message = error_message)
+      super error_message
+      @public_message = if public_error_message.nil?
+                          I18n.t(:"errors.messages.carrierwave_integrity_error")
+                        else
+                          public_error_message
+                        end
+    end
+  end
+
   class InvalidParameter < UploadError; end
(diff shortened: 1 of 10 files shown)


mshibuya:
Sorry for the late response 🙇  

What was in my mind when I wrote [the comment](https://github.com/carrierwaveuploader/carrierwave/pull/2499#issuecomment-778584818) was...

- CarrierWave should raise an exception with enough information to solve the issue for the developers if they are to be blame for the error (like forgetting to install ImageMagick, suppling wrong arguments for ImageMagick or so)
- Otherwise, users are to be blame for the error so CarrierWave returns a validation error to them (As you suggested, this message shouldn't contain implementation-specific information due to security concern. And ideally, the validation error message is actionable one: if the error was caused because they uploaded a non-image file, CarrierWave will report that there was a error because the uploaded file was not an image)

Sorry this may a bit complex, if this explanation doesn't make sense to you please ask for clarification.
I think this is the way we can take to improve security while keeping letting end-users know how to correct when they made a mistake.

akihikodaki:
@mshibuya I don't think it is possible to identify the cause of the error (the server operator or the user) as the error provided by ImageMagick is not structured. If you have an idea for that let me know.
Also, running it without ImageMagick would probably result in a different error though I have not tested.

mshibuya:
I agree that it's not easy, but I guess kind of string matching against the ImageMagick error message could work...
Then let's go step by step. I'll merge #2499 first, then I will try to find the way to implement the friendlier approach above.

I'll close this PR for now, thank you for trying.

akihikodaki:
@mshibuya I don't think such a string is designed as an machine interface. ImageMagick may change (or have changed) such a string without notifying the developer and result in an unexpected failure, potentially in a security-sensitive way. It is somewhat like depending on private methods.

PR #2543, opened 2021-03-15
Add conditional process in README

muratiger:
Conditional version was written, but conditional process was not in README.


mshibuya:
👍

PR #2546, opened 2021-03-18
fix: remove uploaded files if transaction is rolled back

fsateler:
This prevents accumulating stale files when transaction is aborted.

Depends on #2545 , restores functionality added by #2209

### DIFF
--- a/lib/carrierwave/orm/activerecord.rb
+++ b/lib/carrierwave/orm/activerecord.rb
@@ -62,7 +62,9 @@ module CarrierWave
       after_commit :"mark_remove_#{column}_false", :on => :update
 
       after_save :"store_previous_changes_for_#{column}"
+      after_commit :"reset_previous_changes_for_#{column}"
       after_commit :"remove_previously_stored_#{column}", :on => :update
+      after_rollback :"remove_rolled_back_#{column}"
 
       mod = Module.new
       prepend mod
--- a/lib/carrierwave/orm/activerecord.rb
+++ b/lib/carrierwave/orm/activerecord.rb
@@ -56,12 +56,13 @@ module CarrierWave
       validates_processing_of column if uploader_option(column.to_sym, :validate_processing)
       validates_download_of column if uploader_option(column.to_sym, :validate_download)
 
+      after_save :"store_#{column}!"
       before_save :"write_#{column}_identifier"
-      after_save :"store_previous_changes_for_#{column}"
       after_commit :"remove_#{column}!", :on => :destroy
       after_commit :"mark_remove_#{column}_false", :on => :update
+
+      after_save :"store_previous_changes_for_#{column}"
       after_commit :"remove_previously_stored_#{column}", :on => :update
-      after_commit :"store_#{column}!", :on => [:create, :update]
 
       mod = Module.new
       prepend mod
--- a/spec/orm/activerecord_spec.rb
+++ b/spec/orm/activerecord_spec.rb
@@ -780,48 +780,10 @@ describe CarrierWave::ActiveRecord do
       Event.transaction do
         @event.image = stub_file('new.jpeg')
         @event.save
+        expect(File.exist?(public_path('uploads/new.jpeg'))).to be_truthy
         expect(File.exist?(public_path('uploads/old.jpeg'))).to be_truthy
         raise ActiveRecord::Rollback
       end
-      expect(File.exist?(public_path('uploads/old.jpeg'))).to be_truthy
-    end
-  end
-
-  describe "#mount_uploader into transaction" do
-    before do
-      @uploader.version :thumb
-      reset_class("Event")
-      Event.mount_uploader(:image, @uploader)
-      @event = Event.new
-    end
-
-    after do
-      FileUtils.rm_rf(public_path("uploads"))
-    end
-
-    it "should not store file during rollback" do
-      Event.transaction do
-        @event.image = stub_file('new.jpeg')
-        @event.save
-
-        raise ActiveRecord::Rollback
-      end
-
-      expect(File.exist?(public_path('uploads/new.jpeg'))).to be_falsey
-    end
-
-    it "should not change file during rollback" do
-      @event.image = stub_file('old.jpeg')
-      @event.save
-
-      Event.transaction do
-        @event.image = stub_file('new.jpeg')
-        @event.save
-
-        raise ActiveRecord::Rollback
-      end
-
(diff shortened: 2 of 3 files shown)


lib/carrierwave/orm/activerecord.rb
```diff
@@ -56,12 +56,15 @@ def mount_base(column, uploader=nil, options={}, &block)
...
       validates_download_of column if uploader_option(column.to_sym, :validate_download)
 
+      after_save :"store_#{column}!"
       before_save :"write_#{column}_identifier"
-      after_save :"store_previous_changes_for_#{column}"
       after_commit :"remove_#{column}!", :on => :destroy
       after_commit :"mark_remove_#{column}_false", :on => :update
+
+      after_save :"store_previous_changes_for_#{column}"
+      after_commit :"reset_previous_changes_for_#{column}"
```
fsateler, line 65: Honestly, this confuses me. Is the order of callbacks the reverse of add order?

fsateler: [It appears so](https://api.rubyonrails.org/classes/ActiveSupport/Callbacks.html#method-i-run_callbacks): 

> Calls the before and around callbacks in the order they were set, yields the block (if given one), and then runs the after callbacks in reverse order.

Given that this is an after callback, they are executed in reverse.

mshibuya:
These tests need to be added back to ensure that this change has the effect intended in #2209.
https://github.com/carrierwaveuploader/carrierwave/pull/2546/commits/88e3717227e273f1b2caaf08d1fd8620e1462209#diff-48bc92466ff9653e1ae5c5b60134a0c7c631b0d414f0d9b0b4642e84097a650bL790-L828

In addition to that, you need to add tests for the functionality stated in #2544.

fsateler:
> These tests need to be added back to ensure that this change has the effect intended in #2209.
> 88e3717#diff-48bc92466ff9653e1ae5c5b60134a0c7c631b0d414f0d9b0b4642e84097a650bL790-L828

The tests are there. They are reverted in the first commit but re-added in the second.

> In addition to that, you need to add tests for the functionality stated in #2544.

The tests are already there. By reading `public_path` instead of `file_path`, the tests were reading temporary files instead of the definitive one (ie, they were reading the bogus urls).

Do you have a specific test in mind?

spec/orm/activerecord_spec.rb
```diff
@@ -672,7 +672,7 @@ def filename
     end
 
     after do
-      FileUtils.rm_rf(public_path("uploads"))
+      FileUtils.rm_rf(file_path("uploads"))
```
mshibuya, line 675: The change from `public_path` to `file_path`  need to be restored back. As stated in #2209,

> Changes in spec/orm/activerecord_spec.rb:
>
> - Add tests which covers this flows
> - replace cleaning folder from 'file_path' to 'public_path'. Cause file_path is 'spec/fixtures'
>
> and we don't create any 'uploads' folders there. We works with 'spec/public/uploads' folder.

spec/orm/activerecord_spec.rb
```diff
@@ -780,10 +780,12 @@ def filename
       Event.transaction do
         @event.image = stub_file('new.jpeg')
         @event.save
+        expect(File.exist?(public_path('uploads/new.jpeg'))).to be_truthy
```
mshibuya, line 783: These expectations should be added as a separate test case, since they do not match with the subject of this case `it 'should not remove old file if transaction is rollback'`.

mshibuya:
> > These tests need to be added back...
>
> The tests are there.

Sorry I overlooked, please ignore this.

> Do you have a specific test in mind?

These points are stated in #2544, so we should have corresponding test cases to ensure they are resolved by this change.

> For example, reading my_object.file_url.url will return a bogus url if I read before commit.
> For example, if there is a network failure. This causes the transaction to not be rolled back and invalid records being recorded.

spec/orm/activerecord_spec.rb
fsateler: I don't understand this. `file_path` would point to `spec/fixtures/uploads`, and `public_path` points to `spec/public/uploads`. 

So, if I understand correctly, `public_path` is the final destination, and thus is the path to use.

fsateler:
> These points are stated in #2544, so we should have corresponding test cases to ensure they are resolved by this change.
> 
> > For example, reading my_object.file_url.url will return a bogus url if I read before commit.
> > For example, if there is a network failure. This causes the transaction to not be rolled back and invalid records being recorded.

Done.

tomprats:
Would just like to echo that moving this back to a `before_save` is very helpful. I've been getting a mysterious: `OpenURI::HTTPError:  403 Forbidden` in a background job triggered in an `after_commit` ever since I upgraded to 2. What confused me even more was that it always succeeds when retried.

Appreciate the work on this!

spec/orm/activerecord_spec.rb
mshibuya: > file_path would point to spec/fixtures/uploads, and public_path points to spec/public/uploads

That's correct, and we don't create anything in spec/fixtures/uploads. During test execution, files are stored in spec/public/uploads so it need to be cleaned up.

> public_path is the final destination, and thus is the path to use.

That's also true, but you're trying to change from `public_path("uploads")` to `file_path("uploads")`, which results in leaving junks in spec/public/uploads.

fsateler: I don't understand. Was this part wrong before #2209 ? Because this restores the behavior prior to that commit.

mshibuya: > Was this part wrong before #2209 ?

Yes, in that sense #2209 consists of 2 fixes(of course this is not ideal).

Let me illustrate how the issue leads to incorrect result. There's a spec here:

https://github.com/carrierwaveuploader/carrierwave/blob/d50d80eee95054d4e2567b685c5b82a8128146f7/spec/orm/activerecord_spec.rb#L686-L692

and if you comment out the line 689, the spec naturally fails since you have skipped storing files.
But if you use `FileUtils.rm_rf(file_path("uploads"))` instead of `public_path` here:

https://github.com/carrierwaveuploader/carrierwave/blob/d50d80eee95054d4e2567b685c5b82a8128146f7/spec/orm/activerecord_spec.rb#L674-L676

the spec no longer fails because those files are left undeleted from the previous example execution.

fsateler: Ah, I understand now. Thank you for your patience :)

I have updated the PR.

mshibuya:
Thank you for your work 🙌

PR #2550, opened 2021-03-25
Replace mimemagic gem with ruby-magic

pjmartorell:
- Replace `mimemagic` gem with `ruby-magic`
- Some tests are failing, in example:
  - CarrierWave::SanitizedFile#content_type does not allow spoofing of the mime type if the mime type is not detectable
    ```
     expected: "invalid/invalid"
     got: "image/x-mvg"
    ```
    But since the file is a ImageMagick Vector Graphic it seems it should be correct to return `image/x-mvg` instead of `invalid/invalid`
  - CarrierWave::SanitizedFile#content_type does not allow spoofing of the mime type
    ```
     expected: "application/zip"
     got: "application/octet-stream"
    ```
  - CarrierWave::SanitizedFile with a valid File object it should behave like all valid sanitized files #move_to should detect content type correctly using FileMagic when content_type is not set
      ```
     expected:  "invalid/invalid"
     got: "text/plain"
    ```
    But since the file contains the text `this is stuff` it seems it should be correct to return `text/plain` instead of `invalid/invalid`

### DIFF
--- a/carrierwave.gemspec
+++ b/carrierwave.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |s|
   s.add_dependency "activemodel", ">= 5.0.0"
   s.add_dependency "mini_mime", ">= 0.1.3"
   s.add_dependency "image_processing", "~> 1.1"
-  s.add_dependency "mimemagic", ">= 0.3.0"
+  s.add_dependency "ruby-magic", "~> 0.3.1"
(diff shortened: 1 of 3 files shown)


pjmartorell:
Closing this PR in favor of https://github.com/carrierwaveuploader/carrierwave/pull/2551

PR #2551, opened 2021-03-26
Replace mimemagic with marcel gem

### LINKED ISSUE #2548: Update dependency on `mimemagic` to allow non-GPL usage of Carrierwave
The gem `mimemagic` was previously licensed under MIT, but all existing versions were yanked this morning and replaced with new versions licensed under GPL in order to resolve a licensing issue (https://github.com/minad/mimemagic/issues/97). The Rails community is currently working to update or replace this dependency, so that Rails apps are not required to be licensed under the GPL by default (https://github.com/rails/rails/issues/41750).

This issue also affects Carrierwave, because Carrierwave depends on `mimemagic` for the [`mime_magic_content_type` method](https://github.com/carrierwaveuploader/carrierwave/blob/c10b82ed2f7b20cb58772281e3510dc70c410732/lib/carrierwave/sanitized_file.rb#L335). While the Carrierwave dependency does support the new GPL-licensed versions of `mimemagic` (because the dependency is [not locked to a specific version](https://github.com/carrierwaveuploader/carrierwave/blob/5404e5c37f3a86197a8f970a58827f198b586100/carrierwave.gemspec#L28)), it probably makes sense to explore alternatives, as the Rails community is currently doing.

—

UPDATE: We have created a fork (https://github.com/splitwise/carrierwave) that simply removes `mimemagic` from Carrierwave. `mimemagic` was used to add Carrierwave-level mitigation for CVE-2016-3714 ("ImageTragick") in 2.0.0.rc, but ImageTragick can be mitigated in other ways as well, so some applications may not need this level of protection. Removing `mimemagic` also lowers the accuracy of mime type detection, relying solely on filenames to detect file type. I'm unsure whether this is the direction that maintainers would actually like to go in, but the fork may be an appropriate short-term fix for some use cases.

pjmartorell:
- Replace `mimemagic` with `marcel` gem
- Resolves #2548

Waiting for https://github.com/rails/marcel/pull/30 to be merged to add the correct `marcel` version

### DIFF
--- a/carrierwave.gemspec
+++ b/carrierwave.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |s|
   s.add_dependency "activemodel", ">= 5.0.0"
   s.add_dependency "mini_mime", ">= 0.1.3"
   s.add_dependency "image_processing", "~> 1.1"
-  s.add_dependency "mimemagic", ">= 0.3.0"
+  s.add_dependency "marcel", "~> 1.0.0"
   s.add_dependency "addressable", "~> 2.6"
   s.add_dependency "ssrf_filter", "~> 1.0"
   if RUBY_ENGINE == 'jruby'
--- a/lib/carrierwave/sanitized_file.rb
+++ b/lib/carrierwave/sanitized_file.rb
@@ -1,8 +1,7 @@
 require 'pathname'
 require 'active_support/core_ext/string/multibyte'
 require 'mini_mime'
-require 'mimemagic'
-require 'mimemagic/overlay'
+require 'marcel'
 
 module CarrierWave
 
@@ -262,7 +261,7 @@ module CarrierWave
     def content_type
       @content_type ||=
         existing_content_type ||
-        mime_magic_content_type ||
+        marcel_magic_content_type ||
         mini_mime_content_type
     end
 
@@ -329,14 +328,14 @@ module CarrierWave
       end
     end
 
-    def mime_magic_content_type
+    def marcel_magic_content_type
       if path
         type = File.open(path) do |file|
-          MimeMagic.by_magic(file).try(:type)
+          Marcel::Magic.by_magic(file).try(:type)
         end
 
         if type.nil?
-          type = ::MiniMime.lookup_by_filename(path).try(:content_type)
+          type = Marcel::Magic.by_path(file).try(:type)
           type = 'invalid/invalid' unless type.nil? || type.start_with?('text/')
         end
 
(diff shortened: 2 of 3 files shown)


pjmartorell:
@mshibuya or any other maintainer, could you review this PR? I also found a problem installing the dependencies which is making some workflow jobs to fail, regretfully I don't know how to fix it.

pjmartorell:
Pending of the following branches to be merged:
- https://github.com/rails/rails/tree/v6.0.3.5%2Bmimemagic-removal
- https://github.com/rails/rails/tree/v6.1.3%2Bmimemagic-removal
- https://github.com/rails/rails/tree/5-2-stable%2Bmimemagic-removal

dvandersluis:
New versions of rails have been released now which use marcel 1.0. Will this be merged anytime soon?

mejackreed:
Also adding my 👍 to this as upstream mimemagic is not including `mimemagic/overlay` in recent versions causing issues with carrierwave.

https://github.com/mimemagicrb/mimemagic/issues/140

pjmartorell:
I'll try to run tests again and see if they pass now that Rails has released new patched versions

pjmartorell:
All tests passing, who is the maintainer of this gem? 🤷 can we get rid of `mimemagic`?

oakbow:
great work! I am looking forward to seeing this PR merged.
and how about removing mini_mine too?
marcel removed not only mimemagic but mini_mine.
@pjmartorell 

https://github.com/rails/marcel/pull/30/files#diff-89cade48462044ee1b672dc5f4c3ec250fbd29effcd8932096a23c1283c6731fL6

mshibuya:
Following the choice Rails has made looks a steady solution to me. Thanks for the contribution!

unkn0wn-developer:
When will you guys release a new version(2.3.0?) of CarrierWave that uses the new version of Marcel and removes mimemagic? so that we can get rid of mimemagic gem completely in our rails apps after it caused this much of a trouble. 

Mimemagic drastically changed on 0.4.0 and it's very different on 0.3.x. Everything breaks unless it is explicitly added/removed with the correct version, auto-dependency management doesn't work with it etc. That's not desirable. 

So, we are looking forward to use the new CarrierWave with mimemagic removed like Rails did.

dvandersluis:
2.2.1 was released earlier today that removes the mimemagic dependency.

unkn0wn-developer:
Thank you so much guys!!!

heldersantosmoreira:
Are there plans to release a 1.3.3 with the same work?

DmytroVasin:
Are there plans to release a 1.3.3 with the same work? (Up)

mshibuya:
@heldersantosmoreira @DmytroVasin This change is not applicable to 1.x, since it has no dependency for mimemagic.

PR #2552, opened 2021-03-26
Replace mini_mime with marcel

pjmartorell:
- Replaces `mini_mime` with `marcel` since `marcel` has the same functionality 
- Pros:
  - The number of dependencies is reduced (less complexity, less bugs)
  - More consistency since mime types and extensions would be handled by a single gem
  - `marcel` is maintained by Rails core team, which is a warranty of long-term support and development
  - `marcel` identifies 1243 extensions while `mini_mime` identifies 1210
    ```
    irb(main)> Marcel::EXTENSIONS.count
    => 1243

    irb(main)> File.readlines(MiniMime::Configuration.ext_db_path).each do |line|
    irb(main)*     s << line.strip
    irb(main)> end
    irb(main)> File.readlines(MiniMime::Configuration.content_type_db_path).each do |line|
    irb(main)*     s << line.strip
    irb(main)> end
    irb(main)> s.length
    => 1210
    ```
 
- Cons:
  - Introduces the following [breaking change](https://github.com/carrierwaveuploader/carrierwave/blob/70f098c3945c7f82bac8ac1f02b16c5fa314fbae/lib/carrierwave/downloader/remote_file.rb#L35), since some mime types like `image/jpeg` have more than one possible extension (`jpg`, `jpeg`, etc.):

    ```ruby
    Failures:

    1) CarrierWave::Downloader::RemoteFile#original_filename sets file extension based on content-type if missing
       Failure/Error: expect(subject.original_filename).to eq "test.jpeg"
     
         expected: "test.jpeg"
              got: "test.jpg"
     
         (compared using ==)
       # ./spec/downloader/remote_file_spec.rb:58:in `block (3 levels) in <top (required)>'

    2) CarrierWave::Downloader::RemoteFile#original_filename when filename is quoted and empty sets file extension based on content-type if missing
       Failure/Error: expect(subject.original_filename).to eq 'test.jpeg'
     
         expected: "test.jpeg"
              got: "test.jpg"
     
         (compared using ==)
       # ./spec/downloader/remote_file_spec.rb:73:in `block (4 levels) in <top (required)>'

    3) CarrierWave::Downloader::RemoteFile#original_filename when filename is not quoted and empty reads filename correctly
       Failure/Error: expect(subject.original_filename).to eq 'test.jpeg'
     
         expected: "test.jpeg"
              got: "test.jpg"
     
         (compared using ==)
       # ./spec/downloader/remote_file_spec.rb:81:in `block (4 levels) in <top (required)>'
    ```
    - Regarding memory consumption and performance, `mime_magic` has a hash cache of 200 rows and misses are binary-searched from a file (which seems less performant apparently) while `marcel` loads all records in a hash in memory. I don't think there is a noticeable improvement in performance respect what `mime_magic` does, benchmarks should be performed to compare both solutions.

### DIFF
--- a/lib/carrierwave/sanitized_file.rb
+++ b/lib/carrierwave/sanitized_file.rb
@@ -1,6 +1,5 @@
 require 'pathname'
 require 'active_support/core_ext/string/multibyte'
-require 'mini_mime'
 require 'marcel'
 
 module CarrierWave
@@ -261,8 +260,8 @@ module CarrierWave
     def content_type
       @content_type ||=
         existing_content_type ||
-        marcel_magic_content_type ||
-        mini_mime_content_type
+        marcel_magic_by_mime_type ||
+        marcel_magic_by_path
     end
 
     ##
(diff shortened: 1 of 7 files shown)


pjmartorell:
@mshibuya could you take a look at the pros and cons I put in the description and assess if these changes make sense or not?

lib/carrierwave/sanitized_file.rb
```diff
@@ -328,27 +327,27 @@ def existing_content_type
       end
     end
 
-    def marcel_magic_content_type
-      if path
-        type = File.open(path) do |file|
-          Marcel::Magic.by_magic(file).try(:type)
-        end
+    def marcel_magic_by_mime_type
+      return unless path
```
pjmartorell, line 331: I only refactored the `if path` as a guard clause

mshibuya:
Merging in, thanks!

Issue #2559, opened 2021-04-05
"No implicit conversion" error with v2.2.1


robpark:
Using carrierwave-aws 1.5.0, Rails 5.2.5, Ruby 2.6.6.
If I bump carrierwave to 2.2.1, I'm getting errors saving CSV files: `no implicit conversion of CSV into String`. 
Carrierwave 2.2.0 is working fine.

I'm not sure what other information you might find useful, but wanted to bring the issue to your attention.

pjmartorell:
@robpark could you paste the backtrace of the error?

pjmartorell:
@robpark I found a bug and it's addressed in #2562

robpark:
I don't have a backtrace convenient, but will have time tomorrow to reproduce.
I'll also track #2562 and will be happy to test when a patch is out.
Thanks.

robpark:
Here's a backtrace:

```
TypeError:
  no implicit conversion of CSV into String
# /home/rof/cache/bundler/ruby/2.6.0/gems/marcel-1.0.1/lib/marcel/magic.rb:80:in `extname'
# /home/rof/cache/bundler/ruby/2.6.0/gems/marcel-1.0.1/lib/marcel/magic.rb:80:in `by_path'
# /home/rof/cache/bundler/ruby/2.6.0/gems/carrierwave-2.2.1/lib/carrierwave/sanitized_file.rb:338:in `marcel_magic_content_type'
# /home/rof/cache/bundler/ruby/2.6.0/gems/carrierwave-2.2.1/lib/carrierwave/sanitized_file.rb:264:in `content_type'
# /home/rof/cache/bundler/ruby/2.6.0/gems/carrierwave-2.2.1/lib/carrierwave/sanitized_file.rb:221:in `copy_to'
# /home/rof/cache/bundler/ruby/2.6.0/gems/carrierwave-2.2.1/lib/carrierwave/uploader/cache.rb:141:in `cache!'
# /home/rof/cache/bundler/ruby/2.6.0/gems/carrierwave-2.2.1/lib/carrierwave/mounter.rb:63:in `block (2 levels) in cache'
# /home/rof/cache/bundler/ruby/2.6.0/gems/carrierwave-2.2.1/lib/carrierwave/mounter.rb:176:in `handle_error'
# /home/rof/cache/bundler/ruby/2.6.0/gems/carrierwave-2.2.1/lib/carrierwave/mounter.rb:47:in `block in cache'
[...]

PR #2562, opened 2021-04-14
Fix bug when reading content_type from a CSV object linked to a file

pjmartorell:
Resolves https://github.com/carrierwaveuploader/carrierwave/issues/2559

In my previous PR I mistakenly passed the `file` argument to `Marcel::Magic.by_path` instead of `path`. This was not an issue in the majority of cases (`File` objects work properly) but not for a CSV, which raises a `TypeError Exception: no implicit conversion of CSV into String`. Since all tests were passing, this bug passed inadvertently. I added a new test to ensure it does not happen again.

### DIFF
--- a/lib/carrierwave/sanitized_file.rb
+++ b/lib/carrierwave/sanitized_file.rb
@@ -335,7 +335,7 @@ module CarrierWave
         end
 
         if type.nil?
-          type = Marcel::Magic.by_path(file).try(:type)
+          type = Marcel::Magic.by_path(path).try(:type)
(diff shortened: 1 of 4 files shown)


pjmartorell:
@mshibuya could you take a look at this fix?

mshibuya:
Looks great, thanks!

olbrich:
I think this bug also causes failures in the case where the file is empty or the `by_path` method otherwise returns `nil`.  It would be nice to get this released.

PR #2564, opened 2021-04-18
Remove Postgresql and use sqlite3 for testing ActiveRecord

seuros:
I removed Postgres service and used In Memory Sqlite3. This change made the test run much faster as they don't depend on other services.


mshibuya:
This would be handier, thanks!

PR #2566, opened 2021-04-19
Remove Support for Merb

seuros:
Merb was deprecated in 2010. And dont work with current version of ruby


mshibuya:
Thanks for cleaning up!

PR #2568, opened 2021-04-27
Remove redundant each after zip in `remote_urls=`

znz:
The return value changes from an Array to nil.
But it is usually ignored because it is in a setter method.


mshibuya:
👍

PR #2572, opened 2021-05-29
Add Metal Class

seuros:
This class allow the dev to extend different uploader classes  without having the default process/checks included.


mshibuya:
Makes some sense, but could you explain the specific use-case where this is useful?
Also please make the build to pass.

seuros:
I 'm in progress of a big refactoring without breaking old codebase. I will add documentation.

My use case is that i use some uploader that don't require all the callbacks like (remote url, sanitization, versions, or even serialization)

mshibuya:
Stale, please let me know on update.

PR #2577, opened 2021-06-08
Add retry option to download from remote url

tashirosota:
Our project downloads image from imgix by use `remote_image_url`, but imgix occasionally returns a 503 error and download failed.
So I want to support to retry option in carrierwave.
This option will also be useful for other CDNs or storage.

Source: https://aws.amazon.com/premiumsupport/knowledge-center/s3-resolve-503-slowdown-throttling/?nc1=h_ls

### DIFF
--- a/lib/carrierwave/downloader/base.rb
+++ b/lib/carrierwave/downloader/base.rb
@@ -21,6 +21,7 @@ module CarrierWave
       # [remote_headers (Hash)] Request headers
       #
       def download(url, remote_headers = {})
+        @current_download_retry_count = 0
         headers = remote_headers.
           reverse_merge('User-Agent' => "CarrierWave/#{CarrierWave::VERSION}")
         uri = process_uri(url.to_s)
@@ -36,7 +37,13 @@ module CarrierWave
             response.value
           end
         rescue StandardError => e
-          raise CarrierWave::DownloadError, "could not download file: #{e.message}"
+          if @current_download_retry_count < @uploader.download_retry_count
+            @current_download_retry_count += 1
+            sleep 5
+            retry
+          else
+            raise CarrierWave::DownloadError, "could not download file: #{e.message}"
+          end
         end
         CarrierWave::Downloader::RemoteFile.new(response)
       end
--- a/spec/downloader/base_spec.rb
+++ b/spec/downloader/base_spec.rb
@@ -135,6 +135,27 @@ describe CarrierWave::Downloader::Base do
     end
   end
 
+  context 'with download_retry_count' do
+    before do
+      stub_request(:get, uri).to_return({ status: 503 }, { status: 200 })
+    end
+    let(:instance) { described_class.new(uploader) }
+    subject { instance.download uri }
+    context 'when download_retry_count == 0 ' do
+      before { uploader.download_retry_count = 0 }
+      it 'throws an exception' do
+        expect { subject }.to raise_error CarrierWave::DownloadError
+      end
+    end
+
+    context 'when download_retry_count > 0' do
+      before { uploader.download_retry_count = 1 }
+      it 'does not throw an exception' do
+        expect { subject }.not_to raise_error
+      end
+    end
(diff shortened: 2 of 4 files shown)


lib/carrierwave/mounter.rb
```diff
@@ -97,6 +98,15 @@ def remote_urls=(urls)
...
           @uploaders << uploader
+        rescue CarrierWave::DownloadError => e
+          option_download_retry_count = option(:download_retry_count)
+          if option_download_retry_count && current_download_retry_count < option_download_retry_count
+            @current_download_retry_count += 1
+            sleep 5
+            retry
+          else
+            raise e
+          end
```
mshibuya, line 109: Please do not expose a download-specific logic outside of the downloader.
This should be moved to somewhere in [CarrierWave::Downloader::Base](https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/downloader/base.rb).

mshibuya (changes requested):

lib/carrierwave/mounter.rb
tashirosota: Hi @mshibuya!
Thanks for review, and many thanks for carrierwave！
I fixed for move  to CarrierWave::Downloader::Base.

tashirosota: @mshibuya 
Hi, our application require this option.
Is it possible to merge this PR?

lib/carrierwave/downloader/base.rb
```diff
@@ -18,9 +18,11 @@ def initialize(uploader)
       # === Parameters
       #
       # [url (String)] The URL where the remote file is stored
+      # [download_retry_count （Int）] Retry count when download failed
       # [remote_headers (Hash)] Request headers
       #
-      def download(url, remote_headers = {})
+      def download(url, download_retry_count = 0, remote_headers = {})
```
mshibuya, line 24: Please don't insert `download_retry_count` into arguments, this will break existing implementations.

A downloader instance have a reference to an uploader instance, you can just reference `uploader.download_retry_count` inside the downloader.

spec/downloader/base_spec.rb
```diff
@@ -135,6 +135,40 @@
...
+      context 'when download_retry_count > 0' do
+        let(:download_retry_count) { 1 }
+        it { expect { subject }.to raise_error CarrierWave::DownloadError }
+        it do
+          begin
+            subject
+          rescue CarrierWave::DownloadError
+            expect(instance.current_download_retry_count).to eq 1
+          end
+        end
```
mshibuya, line 167: Testing a certain behavior using its internal state (`current_download_retry_count`) is not a good idea. Instead, you can use this API to imitate download attempts which fails initially but successes in the next time.
https://github.com/bblimke/webmock#multiple-responses-for-repeated-requests

spec/downloader/base_spec.rb
```diff
@@ -135,6 +135,40 @@
...
+      stub_request(:get, uri).to_return(status: 503)
+    end
+    let(:instance) { described_class.new(uploader) }
+    subject { instance.download uri, download_retry_count }
+    context 'when failed' do
+      context 'when download_retry_count == 0 ' do
+        let(:download_retry_count) { 0 }
+        it { expect { subject }.to raise_error CarrierWave::DownloadError }
+
+        it do
```
mshibuya, line 149: Don't omit the example name (`it '...'`), unless it can be read naturally using the one-liner syntax.
https://relishapp.com/rspec/rspec-core/docs/subject/one-liner-syntax

mshibuya (changes requested):
Sorry I couldn't follow up, please consider these changes!

tashirosota:
@mshibuya 
Thanks for review, it became very small PR.
I fixed it！

mshibuya (approved):

mshibuya:
Awesome, thanks!

PR #2583, opened 2021-07-07
Fix copying of files > 5GB via aws-fog.

slonopotamus:
This change utilized multipart-upload logic from Fog::AWS::Storage::File#copy that was  added in
https://github.com/fog/fog-aws/commit/e3a9a3521e968436f7d83f11d638b98d504a279b


rsalehi2013:
@mshibuya any chance we can get a new release that includes this patch? thank you

PR #2595, opened 2021-12-27
Require manually features and frameworks.

seuros:
* Paperclip compatibility need to be required manually
   No need to have paperclip compatibility for all the apps
* Test matchers need to be required manually when needed in test/spec_helper.rb
   This don't need to be defined in runtime
* Framework integration needs to be required manually
  I created a Jets namespace (not related to the gem) and it started to break.
* Drop support for sinatra before 1.3.0

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,17 +8,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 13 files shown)


mshibuya:
Please do not put many changes into one commit. At least they should have separate commits (ideally separate PRs).
Also do not including Robocop style changes, it makes tracking diffs hard. Consider making it as another PR.

PR #2612, opened 2022-06-18
suppress some rspec deprecation warning

wonda-tea-coffee:
suppress rspec deprecation warnings

e.g)

```
The implicit block expectation syntax is deprecated, you should pass a block rather than an argument to `expect` to use the provided block expectation matcher or the matcher must implement `supports_value_expectations?`. e.g  `expect { value }.to raise CarrierWave::DownloadError` not `expect(value).to raise CarrierWave::DownloadError`

/home/runner/work/carrierwave/carrierwave/spec/uploader/url_spec.rb:64: warning: already initialized constant FileX
/home/runner/work/carrierwave/carrierwave/spec/uploader/url_spec.rb:64: warning: previous definition of FileX was here
```

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -6,8 +6,6 @@ require 'rspec'
 require 'carrierwave'
 require "webmock/cucumber"
 
-alias :running :lambda
-
 def file_path( *paths )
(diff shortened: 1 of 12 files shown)


mshibuya:
Thank you!

PR #2614, opened 2022-06-27
Set `:content_type` in `copy_options` for GCS.

smnscp:
## Issue Description
Files lose their correct content type when copied on Google Cloud Storage. For example an SVG file with content type "image/svg+xml" becomes "application/octet-stream" when copied to another path on the same bucket.

## Reason
Google Storage API only [considers the `:content_type` option](https://github.com/googleapis/google-api-ruby-client/blob/16c46ef4deb499a390effe9d4edd6a5cbba6cf04/generated/google-apis-storage_v1/lib/google/apis/storage_v1/classes.rb#L1635), so `'Content-Type'` as [defined in `copy_options`](https://github.com/carrierwaveuploader/carrierwave/blob/3cfabaf/lib/carrierwave/storage/fog.rb#L490) is ignored.

## Solution
Set both keys. Beware, I assume that 'Content-Type' is required by AWS but I haven’t checked that.

### DIFF
--- a/lib/carrierwave/storage/fog.rb
+++ b/lib/carrierwave/storage/fog.rb
@@ -487,7 +487,7 @@ module CarrierWave
         def copy_options
           options = {}
           options.merge!(acl_header) if acl_header.present?
-          options['Content-Type'] ||= content_type if content_type
+          options['Content-Type'] = options[:content_type] ||= content_type if content_type
           options.merge(@uploader.fog_attributes)
         end
 
--- a/lib/carrierwave/storage/fog.rb
+++ b/lib/carrierwave/storage/fog.rb
@@ -487,7 +487,7 @@ module CarrierWave
         def copy_options
           options = {}
           options.merge!(acl_header) if acl_header.present?
-          options['Content-Type'] = options[:content_type] ||= content_type if content_type
+          options[fog_provider == "Google" ? :content_type : 'Content-Type'] ||= content_type if content_type
           options.merge(@uploader.fog_attributes)
         end
 
--- a/spec/storage/fog_helper.rb
+++ b/spec/storage/fog_helper.rb
@@ -606,15 +606,14 @@ end
           if @provider == 'AWS'
             expect(@storage.connection).to receive(:copy_object)
                                              .with(anything, anything, anything, anything,
-                                                   { "Content-Type"=>@fog_file.content_type, content_type: @fog_file.content_type, "x-amz-acl"=>"public-read", 'x-amz-server-side-encryption' => 'AES256' }).and_call_original
+                                                   { "Content-Type"=>@fog_file.content_type, "x-amz-acl"=>"public-read", 'x-amz-server-side-encryption' => 'AES256' }).and_call_original
           elsif @provider == 'Google'
             expect(@storage.connection).to receive(:copy_object)
                                              .with(anything, anything, anything, anything,
-                                                   { "Content-Type"=>@fog_file.content_type, content_type: @fog_file.content_type, destination_predefined_acl: "publicRead" }).and_call_original
+                                                   { content_type: @fog_file.content_type, destination_predefined_acl: "publicRead" }).and_call_original
           else
             expect(@storage.connection).to receive(:copy_object)
-                                             .with(anything, anything, anything, anything,
-                                                   { "Content-Type"=>@fog_file.content_type, content_type: @fog_file.content_type }).and_call_original
+                                             .with(anything, anything, anything, anything, { "Content-Type"=>@fog_file.content_type }).and_call_original
           end
 
           @fog_file.copy_to('uploads/new_path.jpg')
@@ -625,16 +624,14 @@ end
         it 'includes acl_header when necessary' do
           if @provider == 'AWS'
             expect(@storage.connection).to receive(:copy_object)
-                                             .with(anything, anything, anything, anything,
-                                                   { "Content-Type"=>@fog_file.content_type, content_type: @fog_file.content_type, "x-amz-acl"=>"public-read" }).and_call_original
+                                             .with(anything, anything, anything, anything, { "Content-Type"=>@fog_file.content_type, "x-amz-acl"=>"public-read" }).and_call_original
           elsif @provider == 'Google'
             expect(@storage.connection).to receive(:copy_object)
                                              .with(anything, anything, anything, anything,
-                                                   { "Content-Type"=>@fog_file.content_type, content_type: @fog_file.content_type, destination_predefined_acl: "publicRead" }).and_call_original
+                                                   { content_type: @fog_file.content_type, destination_predefined_acl: "publicRead" }).and_call_original
           else
             expect(@storage.connection).to receive(:copy_object)
-                                             .with(anything, anything, anything, anything,
-                                                   { "Content-Type"=>@fog_file.content_type, content_type: @fog_file.content_type }).and_call_original
+                                             .with(anything, anything, anything, anything, { "Content-Type"=>@fog_file.content_type }).and_call_original
           end
 
           @fog_file.copy_to('uploads/new_path.jpg')
--- a/spec/storage/fog_helper.rb
+++ b/spec/storage/fog_helper.rb
@@ -606,13 +606,15 @@ end
           if @provider == 'AWS'
             expect(@storage.connection).to receive(:copy_object)
                                              .with(anything, anything, anything, anything,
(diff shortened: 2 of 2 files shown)


thiagofm:
Hi @smnscp thanks for your contribution.

So, naming it `content-type` as a symbol could make it a bit misleading and a bit obscure on our side. I wonder if this is not something as well that could be implemented by the `fog` gem as well.

Also, is the content type of the file being changed by Google Cloud? Or this lies somewhere else?

This scenario is a bit uncommon and corner-case, if we can't find a general and overall good solution, my best guess would be to patch it for your specific use-case. I hope you understand this and please, feel free to contribute more :+1:

smnscp:
Hi @thiagofm, hanks for your reply!

> So, naming it `content-type` as a symbol could make it a bit misleading and would be a bit of a obscure API for Carrierwave.

For one, the method name is `copy_options`, rather than `copy_headers`, so the keys of the returned hash should be named implementation-agnostic. Having symbols as keys is a pretty common choice. For another, the other keys are actually symbols, at least [in case of GCS](https://github.com/carrierwaveuploader/carrierwave/blob/3cfabaf14cfa3c8b922f9774f0e491233e6283fd/lib/carrierwave/storage/fog.rb#L443))—whereas [for AWS there are strings](https://github.com/carrierwaveuploader/carrierwave/blob/3cfabaf14cfa3c8b922f9774f0e491233e6283fd/lib/carrierwave/storage/fog.rb#L496). So, speaking of obscurity—yes, that method could use some light shed on it. 😉

So, let’s go down the rabbit whole. `copy_options` is only called here:
https://github.com/carrierwaveuploader/carrierwave/blob/3cfabaf14cfa3c8b922f9774f0e491233e6283fd/lib/carrierwave/storage/fog.rb#L443

For fog-aws this calls [Fog::AWS::Storage::File#copy](https://github.com/fog/fog-aws/blob/25d720e55c07c8377dd495f5141bcb2ea72f455d/lib/fog/aws/models/storage/file.rb#L145
), then [Fog::AWS::Storage::Real#copy_object](https://github.com/fog/fog-aws/blob/master/lib/fog/aws/requests/storage/copy_object.rb#L30). Here, the `options` param is considered a list of optional headers where keys are expected to be named exactly like the respective headers (in Kebab-Case).

For fog-google, however, we arrive at [Fog::Storage::GoogleJSON::Real#copy_object](https://github.com/fog/fog-google/blob/master/lib/fog/storage/google_json/requests/copy_object.rb#L14). From here, the `options` hash is passed further down to an object initializer where the keys are expected to be proper Ruby symbols (in snake_case).

> I wonder if this is not something as well that could be implemented by the `fog` gem as well, as it is specific to GCS.

I think, you’re right, in a way that the API at Fog’s side should be consistent—which it obviously is not. But instead of relying on the library to have a consistent way of handling the options passed to different implementations Carrierwave should make sure to have the options for all the implementations included.

> Also, is the content type of the file being changed by Google Cloud? Or this lies somewhere else?

Google just doesn’t receive the information about the content type and hence chooses a default one. For some file types (JPEG, PNG) this might still work. But not for all. I don’t know why Google does not just keep the content type of the original file. But that is what happens.

> Anyways, this scenario is a bit uncommon and edge case,

I don’t think so. If Carrierwave passes the content type to Fog::AWS it should also do so for Fog::Google.

> if we can't find a general and overall good solution, my best guess would be to patch it for your specific use-case.

We already have a patch in place. But we hope to get rid of it.

courtsimas:
👀

lib/carrierwave/storage/fog.rb
```diff
@@ -487,7 +487,7 @@ def file
         def copy_options
           options = {}
           options.merge!(acl_header) if acl_header.present?
-          options['Content-Type'] ||= content_type if content_type
+          options['Content-Type'] = options[:content_type] ||= content_type if content_type
```
mshibuya, line 490:
```suggestion
          options[fog_provider == "Google" ? :content_type : 'Content-Type'] ||= content_type if content_type
```

How about using `:content_type` only for fog-google, since it's hard to tell if `:content_type` causes side-effects for other fog providers?

smnscp: LGTM

spec/storage/fog_helper.rb
```diff
@@ -606,13 +606,15 @@ def check_file
           if @provider == 'AWS'
             expect(@storage.connection).to receive(:copy_object)
                                              .with(anything, anything, anything, anything,
-                                                   { "Content-Type"=>@fog_file.content_type, "x-amz-acl"=>"public-read", 'x-amz-server-side-encryption' => 'AES256' }).and_call_original
+                                                   { "Content-Type"=>@fog_file.content_type, content_type: @fog_file.content_type, "x-amz-acl"=>"public-read", 'x-amz-server-side-encryption' => 'AES256' }).and_call_original
```
smnscp, line 609:
```suggestion
                                                   { "Content-Type"=>@fog_file.content_type, "x-amz-acl"=>"public-read", 'x-amz-server-side-encryption' => 'AES256' }).and_call_original
```

spec/storage/fog_helper.rb
```diff
@@ -606,13 +606,15 @@ def check_file
...
           elsif @provider == 'Google'
             expect(@storage.connection).to receive(:copy_object)
-                                             .with(anything, anything, anything, anything, { "Content-Type"=>@fog_file.content_type, destination_predefined_acl: "publicRead" }).and_call_original
+                                             .with(anything, anything, anything, anything,
+                                                   { "Content-Type"=>@fog_file.content_type, content_type: @fog_file.content_type, destination_predefined_acl: "publicRead" }).and_call_original
```
smnscp, line 613:
```suggestion
                                             .with(anything, anything, anything, anything,
                                                   { content_type: @fog_file.content_type, destination_predefined_acl: "publicRead" }).and_call_original
```

spec/storage/fog_helper.rb
```diff
@@ -606,13 +606,15 @@ def check_file
...
           else
             expect(@storage.connection).to receive(:copy_object)
-                                             .with(anything, anything, anything, anything, { "Content-Type"=>@fog_file.content_type }).and_call_original
+                                             .with(anything, anything, anything, anything,
+                                                   { "Content-Type"=>@fog_file.content_type, content_type: @fog_file.content_type }).and_call_original
```
smnscp, line 617:
```suggestion
                                             .with(anything, anything, anything, anything, { "Content-Type"=>@fog_file.content_type }).and_call_original
```

spec/storage/fog_helper.rb
```diff
@@ -623,13 +625,16 @@ def check_file
         it 'includes acl_header when necessary' do
           if @provider == 'AWS'
             expect(@storage.connection).to receive(:copy_object)
-                                             .with(anything, anything, anything, anything, { "Content-Type"=>@fog_file.content_type, "x-amz-acl"=>"public-read" }).and_call_original
+                                             .with(anything, anything, anything, anything,
+                                                   { "Content-Type"=>@fog_file.content_type, content_type: @fog_file.content_type, "x-amz-acl"=>"public-read" }).and_call_original
```
smnscp, line 629:
```suggestion
                                             .with(anything, anything, anything, anything, { "Content-Type"=>@fog_file.content_type, "x-amz-acl"=>"public-read" }).and_call_original
```

spec/storage/fog_helper.rb
```diff
@@ -623,13 +625,16 @@ def check_file
...
           elsif @provider == 'Google'
             expect(@storage.connection).to receive(:copy_object)
-                                             .with(anything, anything, anything, anything, { "Content-Type"=>@fog_file.content_type, destination_predefined_acl: "publicRead" }).and_call_original
+                                             .with(anything, anything, anything, anything,
+                                                   { "Content-Type"=>@fog_file.content_type, content_type: @fog_file.content_type, destination_predefined_acl: "publicRead" }).and_call_original
```
smnscp, line 633:
```suggestion
                                             .with(anything, anything, anything, anything,
                                                   { content_type: @fog_file.content_type, destination_predefined_acl: "publicRead" }).and_call_original
```

smnscp (commented):
If we go with @mshibuya’s [proposal](https://github.com/carrierwaveuploader/carrierwave/pull/2614/files#r923164900) here’s the respective adjustment to the specs.

mshibuya:
Though I agree with the point that this should be handled in fog-google, I think it's somewhat useful as a workaround.
Thank you for the PR!

PR #2620, opened 2022-07-19
Fixed non-existent branch name

wonda-tea-coffee:
rails/sprockets already has no master branch.


wonda-tea-coffee:
Currently, CI fails as follows
https://github.com/carrierwaveuploader/carrierwave/runs/7400610314?check_suite_focus=true

mshibuya:
👍

PR #2624, opened 2022-08-24
Add more details about skip_ssrf_protection config to the Changelog

victmask:
Upgrading carrierwave from any version up to `1.3.1` to a higher version causes for tests to fail complaining about requests not being stubbed properly. 

This PR is adding some notes about it to the Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,7 +121,7 @@ _No changes._
 
 ### Security
 * Fix Code Injection vulnerability in CarrierWave::RMagick (@mshibuya [eb9346df](https://github.com/carrierwaveuploader/carrierwave/commit/eb9346df598a758a5f8c4a338852982fd7f8f6b8), [GHSA-cf3w-g86h-35x4](https://github.com/carrierwaveuploader/carrierwave/security/advisories/GHSA-cf3w-g86h-35x4))
-* Fix SSRF vulnerability in the remote file download feature (@mshibuya [91714add](https://github.com/carrierwaveuploader/carrierwave/commit/91714adda998bc9e8decf5b1f5d260d808761304), [GHSA-fwcm-636p-68r5](https://github.com/carrierwaveuploader/carrierwave/security/advisories/GHSA-fwcm-636p-68r5))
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -121,7 +121,7 @@ _No changes._
 
 ### Security
 * Fix Code Injection vulnerability in CarrierWave::RMagick (@mshibuya [eb9346df](https://github.com/carrierwaveuploader/carrierwave/commit/eb9346df598a758a5f8c4a338852982fd7f8f6b8), [GHSA-cf3w-g86h-35x4](https://github.com/carrierwaveuploader/carrierwave/security/advisories/GHSA-cf3w-g86h-35x4))
-* Fix SSRF vulnerability in the remote file download feature (@mshibuya [91714add](https://github.com/carrierwaveuploader/carrierwave/commit/91714adda998bc9e8decf5b1f5d260d808761304), [GHSA-fwcm-636p-68r5](https://github.com/carrierwaveuploader/carrierwave/security/advisories/GHSA-fwcm-636p-68r5))
+* [BREAKING CHANGE] Fix SSRF vulnerability in the remote file download feature (@mshibuya [91714add](https://github.com/carrierwaveuploader/carrierwave/commit/91714adda998bc9e8decf5b1f5d260d808761304), [GHSA-fwcm-636p-68r5](https://github.com/carrierwaveuploader/carrierwave/security/advisories/GHSA-fwcm-636p-68r5)). It adds a new `skip_ssrf_protection?` config which is false by default. It doesn't allow local requests to be made and breaks tests and requests to private networks ([#2531](https://github.com/carrierwaveuploader/carrierwave/issues/2531))
```
mshibuya, line 124: Adding this instruction is good, but I don't consider this as a `[BREAKING CHANGE]`. Being safe by default is a feature that CarrierWave should have had, but it haven't until then.

victmask: I agree that's this new feature is necessary and important. I believe though it's a breaking change because when upgrading to that version it causes for some components to fail and requires changes to the system.

mshibuya: Please think this way, your app broke because it depended on the unsafe behavior.

Closing this, thank you for the opinion.

victmask: I still think that it should be regarded as a breaking change. A change is still a breaking change when it crashes an existing app. Even if it fixes a security issue. 

Consider these points of view
> It goes without saying that intentional breaking changes should only occur in major versions **with the exception of security fixes** that require users of your library take some action.
https://damieng.com/blog/2021/03/30/breaking-changes/#managing-breaking-changes

> it may happen that a security fix requires a breaking change. 
> In that case one strategy could be to add the more secure way to use the piece of software with additions with the old problematic parts deprecated (and that deprecation and its reason laid out in a clear, easy to find documentation), in which case it is only a MINOR upgrade.
> https://github.com/semver/semver/issues/532#issuecomment-532110917

PR #2629, opened 2022-09-15
Support ssrf_filter 1.1

### LINKED ISSUE #2625: ssrf_filter '1.1.0' introducing a breaking change
Since `ssrf_filter` updated its [version](https://github.com/arkadiyt/ssrf_filter/pull/51/files) from `1.0.8` to `1.1.0`, there is an error being raised:

```
def filename_from_uri
  CGI.unescape(File.basename(uri.path))
end

# bash
undefined method `path' for nil:NilClass

CGI.unescape(File.basename(uri.path))

```

To give more information, the `uri` object above is constructed within the following code snippet:

```
response = SsrfFilter.get(uri, headers: headers) do |req|
  request = req
end
response.uri = request.uri
```

`request.uri` now comes as `nil` from the `SsrfFilter.get` method.

I am not sure whether this issue belongs in here, or if it should be posted to `ssrf_filter`.

BrianHawley:
Includes a backwards compatibility mode for SsrfFilter 1.0,
which is needed for Ruby 2.5 and JRuby 9.2.

Also includes a local version constraint for fog-google for
Ruby 2.5 and JRuby 9.2 because of a known but undeclared
incompatibility in that gem.

[Fixes #2625]

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,6 +36,8 @@ jobs:
     - uses: actions/checkout@v2
     - name: Setup ImageMagick policy
       run: sudo sh -c 'echo '\''<policymap><policy domain="coder" rights="read|write" pattern="PDF" /></policymap>'\'' > /etc/ImageMagick-6/policy.xml'
+    - name: Update package list
+      run: sudo apt update
     - name: Install ghostscript to process PDF
       run: sudo apt-get -y install ghostscript
     - name: Install libvips-dev for Carrierwave::Vips
--- a/Gemfile
+++ b/Gemfile
@@ -3,5 +3,7 @@ source "https://rubygems.org"
 gem "activemodel-serializers-xml"
 gem 'sqlite3', platforms: :ruby
 gem "activerecord-jdbcsqlite3-adapter", platform: [:jruby, :truffleruby]
+# See https://github.com/fog/fog-google/issues/535 for this restriction.
+gem "fog-google", "~> 1.13.0" if RUBY_VERSION.to_f < 2.6
 
 gemspec
--- a/carrierwave.gemspec
+++ b/carrierwave.gemspec
@@ -46,5 +46,4 @@ Gem::Specification.new do |s|
   if RUBY_ENGINE != 'jruby'
     s.add_development_dependency "pry-byebug"
   end
-  s.add_development_dependency "ssrf_filter", "< 1.1.0"
 end
--- a/lib/carrierwave/downloader/base.rb
+++ b/lib/carrierwave/downloader/base.rb
@@ -30,8 +30,12 @@ module CarrierWave
             response = OpenURI.open_uri(process_uri(url.to_s), headers)
           else
             request = nil
-            response = SsrfFilter.get(uri, headers: headers) do |req|
-              request = req
+            if ::SsrfFilter::VERSION.to_f < 1.1
+              response = SsrfFilter.get(uri, headers: headers) do |req|
+                request = req
+              end
+            else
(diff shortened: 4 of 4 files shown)


BrianHawley:
I tried to use a refinement to make the code compatible with old SsrfFilter, but it turns out that Ruby 2.5 isn't able to refine singleton_class modules. Runtime testing works well enough.

BrianHawley:
@mshibuya could you take a look? This fixes ssrf_filter support, and fog-google tests on Ruby 2.5.

lib/carrierwave/downloader/remote_file.rb
```diff
@@ -18,6 +18,8 @@ def initialize(file)
           @headers = file.meta
           @uri = file.base_uri
         end
+      rescue StandardError => e
+        raise CarrierWave::DownloadError, "could not download file: #{e.message}"
```
mshibuya, line 22: Why is this change?

mshibuya:
Thanks for the PR. Ideally this issue should be fixed by ssrf-filter (hence [my PR](https://github.com/arkadiyt/ssrf_filter/pull/57)), but it may take long to be merged.
I think it's good to merge this as a workaround, after passing the review.

lib/carrierwave/downloader/remote_file.rb
BrianHawley: There is another bug in ssrf_filter 1.1 that is not fixed by your PR, which sometimes causes it to fail to retrieve the body (or maybe when it failed to retrieve the body), without letting the Net::HTTPResponse `reading_body` method call `self.body` in its ensure clause, and *not raising an exception*. Then, when `file.body` is called above outside of the `reading_body` method we get that `stream_check` error people were mentioning in the issue.

I don't know what causes this, but I definitely have seen Rollbar reports of this happening on our servers after applying the patch for the `request_proc` problem (I tried both your patch and mine, separately), and it still happened, and I don't know why. Only downgrading ssrf_filter made that problem go away.

Regardless of whether you PR gets merged, this change is definitely needed. Given the tools used, we can't assume that all the exceptions will be raised before you get to this file.

For one thing, if the response of the download was 204, then `file.body` will be nil, and `StringIO.new(file.body)` will raise a TypeError. At the very least, if you don't want to make a distinction between no body and and an empty body, that should be changed to `StringIO.new(file.body.to_s)` so we don't raise an exception on a condition that definitely can happen normally.

If I get the time, I'll try to track down the bug in ssrf_filter that causes these `stream_check` errors in the first place. I already made a PR to net-http to fix the NoMethodError on nil.

mshibuya: Please do not make a workaround. If you think that we need to deal with the situation of '204 No Content', please do so in a separate PR.
Raising `CarrierWave::DownloadError` results in a validation error, but users can't do anything on getting errors like https://github.com/arkadiyt/ssrf_filter/issues/58.

BrianHawley:
@mshibuya I removed the rescue code per request.

Those test failures were test setup infrastructure, not me.

lib/carrierwave/downloader/remote_file.rb
BrianHawley: Removed the rescue code, per request. Made #2633 for the no content responses.

mshibuya:
Merged in, thank you!

nicosierra:
Would it possible to have a new release?

nachiket87:
> Would it possible to have a new release?

s'il vous plait

Bongo1998 (approved):
gracias :)

PR #2633, opened 2022-09-28
Raise DownloadError when no content is returned

### LINKED ISSUE #2632: Download of a url that returns a 204 raises a TypeError; should raise CarrierWave::DownloadError
In lib/carrierwave/downloader/remote_file.rb you have this code:
```ruby
when Net::HTTPResponse
  @file = StringIO.new(file.body)
```

If the response was 204 No Content, then `file.body` will return nil, which `StringIO.new` can't handle.

Given that this url download was intended to get content, and it got no content, it would make sense to explicitly raise a CarrierWave::DownloadError in this case.

BrianHawley:
Have RemoteFile raise CarrierWave::DownloadError when the HTTP response is of a type that has no content, according to the standards. In those cases, response.body returns nil.

There are 8 response types that have no body, 4 of which we can get, so just check whether the body was nil.

Fixes #2632


BrianHawley:
OK, those test failures were test setup infrastructure, not me.

mshibuya:
Thank you!

PR #2640, opened 2022-11-21
Add CI on GitHub for 1.x-stable branch

marcogregorius:
As branch `1.x-stable` doesn't have CI on GitHub, I added the `test` workflow based on `master` branch.

Notable changes:
- Cherrypicked  changes to test against `sqlite3` instead of `postgres` from https://github.com/carrierwaveuploader/carrierwave/pull/2564/commits/f7beff6b1eb9750b0d9f1aa5f370b13410d5b4f7
- Replace `URI.encode` with `URI::DEFAULT_PARSER.escape` for Ruby 3 tests
- Limit `ssrf_filter` gem to < 1.1.0 due to the breaking changes at 1.1.0
- Relax `fog-google` requirement
- Not including `ruby-head` and `jruby-head` as `ruby-head` fails some tests https://github.com/carrierwaveuploader/carrierwave/actions/runs/3514804110/jobs/5889275676 (not sure why?)


mshibuya:
What prevents you from upgrading to a later version of CarrierWave? I don't recommend people to stay on 1.x, without a good reason.

Even if you want to, please don't change Ruby and Rails versions from what's in [.travis.yml](https://github.com/carrierwaveuploader/carrierwave/blob/1.x-stable/.travis.yml). 1.x is meant to be compatible with them.

jacobvosmaer:
Hi @mshibuya, I'm chiming in as a colleague of @marcogregorius at GitLab.

We are trying to upgrade GitLab to Ruby 3.0. GitLab uses CarrierWave so we want to make sure CarrierWave works with Ruby 3.0.

Several years ago we added a series of monkey patches to CarrierWave 1.x in GitLab. You can read about them [here]( https://docs.gitlab.com/ee/development/uploads/working_with_uploads.html#gitlab-modifications-to-carrierwave). These modifications make it more difficult for us to upgrade to 2.x. Additionally, we have some [concerns](https://gitlab.com/gitlab-org/gitlab/-/issues/216067#note_593408556) about different use of ActiveRecord callbacks in CarrierWave 2.x so we may want to wait for CarrierWave 3.x.

Because of the large number of Ruby gems GitLab depends on, upgrading to Ruby 3.0 is proving to be a lot of work. We would like to avoid making GitLab's Ruby 3.0 upgrade dependent on resolving our (self-created) CarrierWave challenges. What we would need for that is a 1.x version of CarrierWave that is known to work with Ruby 3.

I would completely understand if you do not want to make these kind of changes to the `1.x-stable` branch. At the same time, the changes required for Ruby 3 itself seem minor minor that we thought we should at least ask if we can address them here in this repo.

@mshibuya are you open to having a CarrierWave 1.x release with Ruby 3 support?

mshibuya:
Adding Ruby 3.0 support to 1.x-stable is OK, as long as you maintain the compatibility with Ruby >= 2.2 and Rails >= 4.0. I can cut a new gem release for 1.x if you need it.

And I have one proposal with your 3.0 migration plan.
I don't recommend having monkey-patches since it will make upgrade harder, like you're facing. Are you interested in stopping monkey-patching and integrating features you need to the official repo?
Is the feature you need something similar to [carrierwave_direct](https://github.com/dwilkie/carrierwave_direct)?

jacobvosmaer:
> I don't recommend having monkey-patches since it will make upgrade harder, like you're facing. Are you interested in stopping monkey-patching and integrating features you need to the official repo?

This is a very reasonable question and I appreciate the invitation. I really struggle deciding how to answer this though.

We are not happy with the monkey patches but they work. At this time it is not clear which of the challenges we have with our current solution are most important to address. Being stuck on CarrierWave 1.x is just one of the problems.

Having said that, I want to say again I appreciate the invitation to work inside the CarrierWave code base instead of around it.

marcogregorius:
> Adding Ruby 3.0 support to 1.x-stable is OK, as long as you maintain the compatibility with Ruby >= 2.2 and Rails >= 4.0. I can cut a new gem release for 1.x if you need it.

Thanks @mshibuya , that would be much appreciated!

However, I tried adding Ruby >= 2.2 and Rails >= 4.0 and changed `pg` to `sqlite3` adapter, but the CI failed only for [Rails 4.x](https://github.com/carrierwaveuploader/carrierwave/actions/runs/3564491518/jobs/5988509688). I think there is an issue with the `sqlite3` version and ActiveRecord 4.x with handling JSON column. Do you have any suggestion on how to handle this?

Alternatively, I also tried using the original `postgres` as the `1.x-stable` branch originally had, in this branch https://github.com/marcogregorius/carrierwave/pull/2.  However, I ran into different errors:
- [Ruby 2.2 & 2.3 would throw segfault](https://github.com/marcogregorius/carrierwave/actions/runs/3549202404/jobs/5961299112)  if [host is included in dbconfig](https://github.com/marcogregorius/carrierwave/commit/c1eea2c31cfde52279c56f7f9a452da3343f3aea)
- On the other hand, GitHub actions require the app to specify the host as `127.0.0.1` otherwise the app couldn’t find the PG DB as in [https://github.com/marcogreg](https://github.com/marcogregorius/carrierwave/actions/runs/3563550002/jobs/5986460629).

mshibuya:
OK I can handle the rest. Thank you!

mshibuya:
@jacobvosmaer @marcogregorius Can I proceed with creating a new 1.x release, or do you need some more changes?

jacobvosmaer:
@mshibuya thanks for asking. I see we now have [green CI](https://github.com/carrierwaveuploader/carrierwave/actions/runs/3805565190/jobs/6473603131) on the `1.x-stable` branch, including jobs with Ruby 3.0, which is what we were looking for.

You can go ahead with the release!

mshibuya:
Released!
https://rubygems.org/gems/carrierwave/versions/1.3.3

marcogregorius:
@mshibuya Thanks for the release! We have [bumped GitLab](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/111295) to use the new version since. Again, thanks for the effort!

PR #2641, opened 2022-11-21
Local storage clean_cache! fix for off-topic files

BrianHawley:
Given that it is not advised to put other stuff in the same directory as CarrierWave's cache, this fix makes clean_cache! not break if you do.

Based on the similar fix for the Fog storage clean_cache!.


mshibuya:
Makes sense, thanks!

PR #2642, opened 2022-12-05
Enable to override configurations by 'false' on a per-uploader basis

### LINKED ISSUE #2417: fog_public is not respected in individual uploader
After upgrade to version 2 it looks like for_public is not respected anymore. File is uploaded as private.

```ruby
def fog_public
    true
end
```

megane42:
## motivation

- fixes #2417
- Currently, it's not possible to set `false` to some boolean values (e.g. `fog_public`) via in-class configuring.
- There are some workarounds already, but it would be more convenient and intuitive.
    - [How to: Define different storage configuration for each Uploader.](https://github.com/carrierwaveuploader/carrierwave/wiki/How-to:-Define-different-storage-configuration-for-each-Uploader.)
    - [How to: Secure uploaded file in S3 and make it only accessible by supplying secret token?](https://github.com/carrierwaveuploader/carrierwave/wiki/How-to:-Secure-uploaded-file-in-S3-and-make-it-only-accessible-by-supplying-secret-token%3F)

## step to reproduce

In `config/initializers/carrierwave.rb` :

```ruby
  config.fog_public = true
  ...
end
```

and `avatar_uploader.rb` :

```ruby
class AvatarUploader < CarrierWave::Uploader::Base
  fog_public false
end
```

## expected behavior

```ruby
AvatarUploader.fog_public # => false
```

## actual behavior

```ruby
AvatarUploader.fog_public # => true
```

### DIFF
--- a/lib/carrierwave/uploader/configuration.rb
+++ b/lib/carrierwave/uploader/configuration.rb
@@ -124,7 +124,7 @@ module CarrierWave
             @#{name} = nil
 
             def self.#{name}(value=nil)
-              @#{name} = value if value
+              @#{name} = value unless value.nil?
               return @#{name} if self.object_id == #{self.object_id} || defined?(@#{name})
               name = superclass.#{name}
               return nil if name.nil? && !instance_variable_defined?(:@#{name})
--- a/spec/uploader/configuration_spec.rb
+++ b/spec/uploader/configuration_spec.rb
@@ -116,11 +116,11 @@ describe CarrierWave::Uploader::Base do
     end
 
     it "adds a convenient in-class setter" do
-      uploader_class.foo_bar('bar')
-      expect(uploader_class.foo_bar).to eq('bar')
+      expect(uploader_class.foo_bar).to eq('foo')
+    end
 
-      uploader_class.foo_bar(false)
-      expect(uploader_class.foo_bar).to eq(false)
+    it "adds a convenient in-class setter which can assign 'false' as a value" do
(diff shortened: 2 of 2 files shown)


lib/carrierwave/uploader/configuration.rb
```diff
@@ -124,7 +124,7 @@ def add_config(name)
             @#{name} = nil
 
             def self.#{name}(value=nil)
-              @#{name} = value if value
+              @#{name} = value unless value.nil?
```
megane42, line 127: This assignment is skipped if a given `value` is `false`.

spec/uploader/configuration_spec.rb
```diff
@@ -116,7 +116,11 @@
     end
 
     it "adds a convenient in-class setter" do
-      expect(uploader_class.foo_bar).to eq('foo')
+      uploader_class.foo_bar('bar')
+      expect(uploader_class.foo_bar).to eq('bar')
+
+      uploader_class.foo_bar(false)
+      expect(uploader_class.foo_bar).to eq(false)
```
mshibuya, line 123: Please create a separate test case for this part. The implementation itself looks good!

megane42: Sorry, that's fixed :) bcc3bb0
Should I squash commits?

mshibuya: That's not necessary!

mshibuya:
Merged in, thanks!

PR #2645, opened 2023-01-04
Fix dup with multiple mounters or not cached

### LINKED ISSUE #1962: Support for ActiveRecord dup method
It would be great if carrierwave would be able to fully support ActiveRecords dup method. For example,

if I have the following model

``` ruby
class ExampleModel < ActiveRecord::Base
  mount_uploader :file, FileUploader
end
```

then i'd be able to do the following

``` ruby
current = ExampleModel.find 1
new_model = current.dup
new_model.save
```

And the file would be copied and saved over to the new model.

BrianHawley:
- Don't call `each` if `@_mounters` is unitialized or nil. It will be nil for calls to later mounters, or if not referenced or saved.
- Don't interpolate `column` when it should be a local variable.
- Don't define global functions in specs. Make them local to the block.
- Cover more missing scenarios.
- Fix some rubocop failures.

[Fixes #1962]

### DIFF
--- a/lib/carrierwave/orm/activerecord.rb
+++ b/lib/carrierwave/orm/activerecord.rb
@@ -93,12 +93,10 @@ module CarrierWave
 
         # Reset cached mounter on record dup
         def initialize_dup(other)
-          old_mounters = @_mounters
-          @_mounters = nil
+          old_uploaders = _mounter(:"#{column}").uploaders
+          @_mounters[:"#{column}"] = nil
           super
-          old_mounters.each do |column, mounter|
-            _mounter(:#{column}).cache(mounter.uploaders)
-          end
+          _mounter(:"#{column}").cache(old_uploaders)
         end
       RUBY
     end
--- a/spec/orm/activerecord_spec.rb
+++ b/spec/orm/activerecord_spec.rb
@@ -1,26 +1,26 @@
 require 'spec_helper'
 require 'support/activerecord'
 
-def create_table(name)
-  ActiveRecord::Base.connection.create_table(name, force: true) do |t|
-    t.column :image, :string
-    t.column :images, :json
-    t.column :textfile, :string
-    t.column :textfiles, :json
-    t.column :foo, :string
-  end
-end
-
-def drop_table(name)
-  ActiveRecord::Base.connection.drop_table(name)
-end
-
-def reset_class(class_name)
-  Object.send(:remove_const, class_name) rescue nil
-  Object.const_set(class_name, Class.new(ActiveRecord::Base))
-end
-
 describe CarrierWave::ActiveRecord do
+  def create_table(name)
+    ActiveRecord::Base.connection.create_table(name, force: true) do |t|
+      t.column :image, :string
+      t.column :images, :json
+      t.column :textfile, :string
+      t.column :textfiles, :json
+      t.column :foo, :string
+    end
+  end
+
+  def drop_table(name)
+    ActiveRecord::Base.connection.drop_table(name)
+  end
+
+  def reset_class(class_name)
+    Object.send(:remove_const, class_name) rescue nil
+    Object.const_set(class_name, Class.new(ActiveRecord::Base))
+  end
+
   before(:all) { create_table("events") }
   after(:all) { drop_table("events") }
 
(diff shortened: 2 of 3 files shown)


mshibuya:
Thanks for the PR. Where's the fix for this point?

> The initialize_dup is defined in the wrong context, as it refers to the overall mounters collection for the object, but is defined in a per-mounter module.

https://github.com/carrierwaveuploader/carrierwave/issues/1962#issuecomment-1370308875

spec/orm/activerecord_spec.rb
```diff
@@ -1,26 +1,26 @@
...
 require 'support/activerecord'
 
-def create_table(name)
-  ActiveRecord::Base.connection.create_table(name, force: true) do |t|
-    t.column :image, :string
-    t.column :images, :json
-    t.column :textfile, :string
-    t.column :textfiles, :json
-    t.column :foo, :string
+describe CarrierWave::ActiveRecord, :aggregate_failures do
```
mshibuya, line 4: Could you remove this `:aggregate_failures` ? While it can be useful for local development, I don't see any benefit for permanently enabling this.

BrianHawley: I could remove it, but there's no harm in keeping it, as it's also useful in CI. It gives better errors when things fail, and has no overhead in the success case.

lib/carrierwave/orm/activerecord.rb
```diff
@@ -93,11 +93,11 @@ def reload(*)
...
         # Reset cached mounter on record dup
         def initialize_dup(other)
-          old_mounters = @_mounters
+          old_mounters = @_mounters if instance_variable_defined?(:@_mounters)
           @_mounters = nil
           super
-          old_mounters.each do |column, mounter|
-            _mounter(:#{column}).cache(mounter.uploaders)
+          old_mounters&.each do |column, mounter|
+            _mounter(column).cache(mounter.uploaders)
```
BrianHawley, line 100: Weirdly enough, these two lines have the side effect of fixing the method being defined in the wrong context.

You don't want this method defined unless you have at least one mount field in the model, so it should be added by this method builder somehow. The problem is, the method gets redefined for every mount field, and these two lines apply to all fields and you only want them to run once.

I could have tried to make another module for this and `reload` that only gets included once, and then check whether it's there before including it again. But, that would add a module name and that inclusion check is a little difficult to do compatibly across the Ruby versions supported.

However, I noticed that if this method is defined and called repeatedly in the same inheritance chain, `@_mounters` is nil already for every later call, due to it having been assigned nil in the earlier call. When `old_mounters` is nil, `old_mounters&.each` doesn't do anything. This means that the later calls are safe to run because they don't do anything significant that can't be repeated.

This is the same trick that makes the `reload` method above safe to repeat, as it only assigns a nil to the same variable once per field.

BrianHawley:
@mshibuya I've used this fix in my app at work (to fix after updating carrierwave yesterday) and it solves the issues I ran into: NoMethodError on nil, fortunately caught by our test suite. I fixed it there right before making the PR here.

spec/processing/rmagick_spec.rb
```diff
@@ -235,7 +235,9 @@
 
     it 'does not allow invocation of non-public methods' do
       Kernel.module_eval do
-        private def foo=(value); raise; end
+      private
+
+        def foo=(value); raise; end
```
BrianHawley, line 240: This was just to get a fully passing rubocop run.

lib/carrierwave/orm/activerecord.rb
```diff
@@ -93,11 +93,11 @@ def reload(*)
 
         # Reset cached mounter on record dup
         def initialize_dup(other)
-          old_mounters = @_mounters
+          old_mounters = @_mounters if instance_variable_defined?(:@_mounters)
           @_mounters = nil
```
mshibuya, line 97:
```suggestion
          old_uploaders = _mounter(:"#{column}").uploaders
          @_mounters[:"#{column}"] = nil
```

lib/carrierwave/orm/activerecord.rb
```diff
@@ -93,11 +93,11 @@ def reload(*)
...
           super
-          old_mounters.each do |column, mounter|
-            _mounter(:#{column}).cache(mounter.uploaders)
+          old_mounters&.each do |column, mounter|
+            _mounter(column).cache(mounter.uploaders)
           end
```
mshibuya, line 101:
```suggestion
          _mounter(:"#{column}").cache(old_uploaders)
```

mshibuya (commented):
If we were to solve the 'wrong context' issue, the change will look like this. What do you think?

BrianHawley:
> If we were to solve the 'wrong context' issue, the change will look like this. What do you think?

Sure, that will work. Updated with your changes, @mshibuya.

mshibuya:
Thank you!

PR #2646, opened 2023-01-04
Support to configure download retry wait time

tricknotes:
https://github.com/carrierwaveuploader/carrierwave/pull/2577 is good change that I wanted.
I'll add a config for download retry wait time.

Since the appropriate wait time depends on the each of the remote server, it is better to be able to configure the wait time on the application side.


mshibuya:
Nice feature 👍

PR #2647, opened 2023-01-05
Update CHANGELOG.md for v2.2.3

tricknotes:
This change comes from baf5df7d9acea95f46eaea456743241ef3d644f5. This commit exists only 2.x-stable blanch, does not exist in master blanch.


tricknotes:
Currently, rails main is broken. Ref: https://github.com/rails/rails/pull/46897
Other test cased are passed.
I think there are no problem on this PR.

mshibuya:
Thanks for the PR, but maintaining the changelog in multiple places is costly. Placing the link for old version changelog is better, like [Rails](https://github.com/rails/rails/blob/main/activesupport/CHANGELOG.md) does.

PR #2653, opened 2023-01-23
Bugfix: fix Kernel#=~ removal in Ruby 3.2

aubinlrx:
Since the release of Ruby 3.2 `Kernel#=~` has been removed. Which that `IntegerClass`, `FalseClass` doesn't respond to `def =~` anymore. 

Screenshot 2023-01-23 at 16 10 18
Screenshot 2023-01-23 at 16 10 09

Since the String class still respond to it we could infer `remove` to a string and keep the same check than before. I'm open a more complicated solution which check all types: `Numeric`, `FalseClass` and `String`

### DIFF
--- a/lib/carrierwave/mounter.rb
+++ b/lib/carrierwave/mounter.rb
@@ -116,7 +116,7 @@ module CarrierWave
     end
 
     def remove?
-      remove.present? && (remove == true || remove !~ /\A0|false$\z/)
+      remove.present? && (remove == true || remove.to_s !~ /\A0|false$\z/)
     end
 
     def remove!


mshibuya:
Could you add a test case, so the Ruby 3.2 build will fail without this fix?

lib/carrierwave/mounter.rb
```diff
@@ -116,7 +116,7 @@ def blank?
     end
 
     def remove?
-      remove.present? && (remove == true || remove !~ /\A0|false$\z/)
+      remove.present? && (remove == true || remove.to_s !~ /\A0|false$\z/)
```
mshibuya, line 119:
```suggestion
      remove.present? && (remove.to_s !~ /\A0|false$\z/)
```

We can drop `remove == true` , right?

PR #2657, opened 2023-02-21
Better error message for restrictive imagemagick security policies

TomK32:
I stumbled across this problem when I moved the development on my local machine into a docker container and the original error message pushing the blame on my PDF wasn't helpful at all. I do understand the change a few years ago to not show any details from the original error message, but this was painful. https://github.com/carrierwaveuploader/carrierwave/commit/7af7db3636f79a811e0f4f5da32de50042994493

Some info about the imagemagick security policy can be found here https://imagemagick.org/script/security-policy.php and in this blogpost https://alexvanderbist.com/2018/fixing-imagick-error-unauthorized/

The error message I added will point anyone into the right direction. I had a bit of a trouble with the test, hope it is good enough

### DIFF
--- a/lib/carrierwave/locale/en.yml
+++ b/lib/carrierwave/locale/en.yml
@@ -9,5 +9,6 @@ en:
       content_type_allowlist_error: "You are not allowed to upload %{content_type} files, allowed types: %{allowed_types}"
       content_type_denylist_error: "You are not allowed to upload %{content_type} files"
       processing_error: "Failed to manipulate, maybe it is not an image?"
+      rmagick_error_security_policy: "Failed to manipulate due to imagemagick's security policy for this filetype"
       min_size_error: "File size should be greater than %{min_size}"
       max_size_error: "File size should be less than %{max_size}"
--- a/lib/carrierwave/processing/rmagick.rb
+++ b/lib/carrierwave/processing/rmagick.rb
@@ -366,8 +366,12 @@ module CarrierWave
       end
 
       destroy_image(frames)
-    rescue ::Magick::ImageMagickError
-      raise CarrierWave::ProcessingError, I18n.translate(:"errors.messages.processing_error")
+    rescue ::Magick::ImageMagickError => e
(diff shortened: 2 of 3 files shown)


lib/carrierwave/locale/en.yml
```diff
@@ -9,5 +9,6 @@ en:
       content_type_allowlist_error: "You are not allowed to upload %{content_type} files, allowed types: %{allowed_types}"
       content_type_denylist_error: "You are not allowed to upload %{content_type} files"
       processing_error: "Failed to manipulate, maybe it is not an image?"
+      rmagick_error_security_policy: "Failed to manipulate due to imagemagick's security policy for this filetype"
```
mshibuya, line 12: How about an error message like this?

```suggestion
      security_policy_error: "Uploading this format of files is not allowed due to security reasons. Try different one"
```

This will be given to end-users as a validation message, so a word like `imagemagick` is useless to them.

TomK32: I'd prefer it if that error message would help the user to find the cause of it as fast as possible. "Imagemagick" and "security policy" are must-have keywords for this.

TomK32: And "uploading this format of files is no allowed" is misleading because such cases should be covered by the code at an earlier point.

mshibuya: > such cases should be covered by the code at an earlier point

I don't think so. Having a specific file extension or content type and actually being in some file format is different things.

Then how about re-raising the original exception? In that way you can know something happened with ImageMagick. I can't merge this PR as it is, since it goes against the intention of #2499.

mshibuya: I'm not interested in your _wasted_ hours. I'm maintaining this library for free, can I call this conversation waste of time then?
Let's focus on how to make this library better, please.

My point is that showing the word `imagemagick` to end-users as the validation message is useless and confusing, as stated in #2499. What's useful is to let end-users know what to do next. What do you want them to do? Try another file, or consult with the administrator?

mshibuya:
Stale, please let me know on update.

PR #2660, opened 2023-03-15
Use helper method to access fog_provider

joshuamsager:
Replaces instances of: `@uploader.fog_credentials[:provider]` with it's helper method: `fog_provider`

This proves to be helpful for me because I am trying to build my own Fog provider. In many scenarios, my provider behaves like AWS and in some scenarios it behaves like Google.

Having the ability to monkeypatch a single method: `CarrierWave::Storage::Fog#fog_provider` cleans up my implementation greatly.


mshibuya:
I think it's better, thank you!

PR #2663, opened 2023-05-02
Extend support of 0.11.x with ruby 3.2

kimyu92:
Due to tech debt, we still cannot move directly to v1:

I would like to backport some changes to prolong 0.11.x
- https://github.com/carrierwaveuploader/carrierwave/pull/2653
- Backport https://github.com/carrierwaveuploader/carrierwave/commit/fc0edd8866a49625aeb928f4c8ad2b5c014f6544 to replace File.exists? with File.exist? [deprecated](https://bugs.ruby-lang.org/issues/9041#note-1) since ruby 2.2


mshibuya:
But 0.11.x is too old. What's preventing you from the upgrade?

kimyu92:
Thanks for checking. We had built quite complicated around legacy carrierwave but we figured out a workaround.

Issue #2148, opened 2017-03-27
Carrierwave won't generate URL for conditional versions


bjedrocha:
I'm using carrierwave's [conditional versioning](https://github.com/carrierwaveuploader/carrierwave#conditional-versions) to perform version processing in a background job. Here is my uploader class

```
class LogoUploader < CarrierWave::Uploader::Base
  include CarrierWave::MiniMagick

  attr_writer :process_delayed

  def process_delayed?(_img = nil)
    !!@process_delayed
  end

  # directory files are stored when running in development
  def store_dir
    "uploads/#{model.class.to_s.underscore}/#{mounted_as}/#{model.id}"
  end

  version :optimized, if: :process_delayed? do
    process resize_to_limit: [nil, 200]
  end
end
```

In the background job, I simply set `model.logo.process_delayed = true` prior to running `model.logo.recreate_versions! :optimized`. This works in that it creates my optimized version but then if I try `model.logo.optimized.url` it simply returns `nil`. Interestingly enough, if I just remove the condition from the `version` block, the version url then works. Thus it appears that conditions on versions get re-evaluated when generating the URL but for the life of me I can't find what in the code determines this. Any help would be appreciated.

MGPalmer:
I'm having the same issue :(

MGPalmer:
If I change this line:
https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/uploader/versions.rb#L290
to use "versions" instead of "active_versions", I get the URLs back as it was before.

I'm reluctant to use it for me, though, because this behaviour is explicitly specced...

dnswus:
I'm also having this issue on `1.0` and `1.1`. Have to switch back to `0.11` to make it work.

piotrze:
+1

ahmedrayhan:
is anyone still facing this issue?

MGPalmer:
Sorry, I don't remember if we fixed this back then...

victmask:
it seems to be done intentionally with https://github.com/carrierwaveuploader/carrierwave/pull/1351

mshibuya:
This is how versions are supposed to work.

To achieve what you want, you'll need to persist the processing status into the model. The pseudo-code will look like:

```ruby
class LogoUploader < CarrierWave::Uploader::Base
  include CarrierWave::MiniMagick

  def processed?(_img = nil)
    model.logo_processed
  end

  # directory files are stored when running in development
  def store_dir
[...]

Issue #2132, opened 2017-03-03
Conditional versions checking everytime uploader called?


adysson:
Hi,
after update of Carrierwave from 0.6.2 to 1.0.0, our app started to be painfully slow.
We use conditional versions in an uploader i.e. `Object#logo`. By method `is_big_enough?`, it checks source image resolution (using mini_magick gem) before versions `:medium` and `:large` are created - to prevent increasing of image resolution i.e. from 200x200px to 400x400px.
There are other versions without conditionals: `:small` and `:semi_small`
In Carrierwave v. 0.6.2, if I call `object.logo.small.url`, conditional method `is_big_enough?` is not called.
In Carrierwave v. 1.0.0, conditional method `is_big_enough?` seems to be called everytime.

In a big page with 100+ calls to various instances of the uploader, loading of page starts to be painfully slow, because `is_big_enough?` uses mini_magick which touches HDD/SDD.

Is this desired behavior because of some reason or it happens by mistake?
Thank you

francisneedham:
same problem here

marcek:
Have the same problem currently using version 2.2.2

jasonfine:
Bumping, this needs input from the authors.

mshibuya:
The change itself came from #1351, but makes some sense.
So what I recommend is to [get dimensions of the file and store it in some model columns on processing](https://github.com/carrierwaveuploader/carrierwave/wiki/How-to%3A-Get-image-dimensions), so the dimensions can be checked with little overhead later.

PR #2669, opened 2023-07-04
lately resolve active versions when retrieving

### LINKED ISSUE #2461: Long time delay associated with head requests to Amazon S3
Hello. We use carrierwave to handle attachments stored by Amazon S3. Files in bucket are private. Our API sometimes has to return data about many attachments (100 for instance). In this endpoints we're observing a big delay in response (the more attachments, the longer response time). We've used New Relic to investigate what affects this time, and turns out that HEAD requests are performing one by one (not simultaneously):

image

Moreover, this causes high CPU utilization (sometimes over 50%) - the more S3 requests, the more CPU utilization.
Has anyone encountered a similar problem?

rajyan:
I believe this fixes https://github.com/carrierwaveuploader/carrierwave/issues/2461 https://github.com/carrierwaveuploader/carrierwave/issues/2148

~~https://github.com/carrierwaveuploader/carrierwave/issues/2132~~
 
Minimum reproducible example and details are in
https://github.com/carrierwaveuploader/carrierwave/issues/2461#issuecomment-1620799178

### DIFF
--- a/lib/carrierwave/uploader/versions.rb
+++ b/lib/carrierwave/uploader/versions.rb
@@ -133,7 +133,7 @@ module CarrierWave
 
           class_eval <<-RUBY, __FILE__, __LINE__ + 1
             def #{name}
-              versions[:#{name}]
+              version_exists?(:#{name}) ? versions[:#{name}] : versions[:#{name}].class.new(model, mounted_as)
             end
           RUBY
 
@@ -325,11 +325,11 @@ module CarrierWave
       end
 
       def retrieve_versions_from_cache!(cache_name)
-        active_versions.each { |name, v| v.retrieve_from_cache!(cache_name) }
+        versions.each { |name, v| v.retrieve_from_cache!(cache_name) }
       end
 
       def retrieve_versions_from_store!(identifier)
-        active_versions.each { |name, v| v.retrieve_from_store!(identifier) }
+        versions.each { |name, v| v.retrieve_from_store!(identifier) }
       end
 
     end # Versions
--- a/spec/uploader/versions_spec.rb
+++ b/spec/uploader/versions_spec.rb
@@ -449,6 +449,13 @@ describe CarrierWave::Uploader do
           expect(@uploader.thumb).to be_present
           expect(@uploader.preview).to be_blank
         end
+
+        it "should evaluate the condition even the version is not called" do
+          @uploader_class.version(:preview, if: lambda{|record, args| record.false?(args[:file])})
+          expect(@uploader).to receive(:false?).at_least(:once).and_return(false)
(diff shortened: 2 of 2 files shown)


rajyan:
I noticed that this PR might also fix https://github.com/carrierwaveuploader/carrierwave/issues/2148 https://github.com/carrierwaveuploader/carrierwave/issues/2132
without breaking the purpose of https://github.com/carrierwaveuploader/carrierwave/pull/1351
I would add tests for them if necessary.

lib/carrierwave/uploader/versions.rb
```diff
@@ -325,11 +325,11 @@ def remove_versions!
...
 
       def retrieve_versions_from_cache!(cache_name)
-        active_versions.each { |name, v| v.retrieve_from_cache!(cache_name) }
+        versions.each { |name, v| v.retrieve_from_cache!(cache_name) }
       end
 
       def retrieve_versions_from_store!(identifier)
-        active_versions.each { |name, v| v.retrieve_from_store!(identifier) }
+        versions.each { |name, v| v.retrieve_from_store!(identifier) }
       end
```
rajyan, line 334: We don't need to evaluate the condition here, like before the changes in https://github.com/carrierwaveuploader/carrierwave/pull/1351 .
We can keep all versions retrieved internally, and only evaluate them when users accesses.

lib/carrierwave/uploader/versions.rb
```diff
@@ -133,7 +133,7 @@ def version(name, options = {}, &block)
 
           class_eval <<-RUBY, __FILE__, __LINE__ + 1
             def #{name}
-              versions[:#{name}]
+              version_exists?(:#{name}) ? versions[:#{name}] : versions[:#{name}].class.new(model, mounted_as)
```
rajyan, line 136: Lately resolving the condition here. Which would make it more consistent with the current state (active or not) of the version (which fixes https://github.com/carrierwaveuploader/carrierwave/issues/2148 too).

I think it is better to return nil for non-active versions, but returning a blank uploader for backwards compatibility.

rajyan:
@mshibuya 

I added some comments to make the change clearer. Hope this makes sense :smile:

rajyan:
Looking at https://github.com/carrierwaveuploader/carrierwave/issues/2132
carefully, this PR didn’t solve the issue.

rajyan:
Closing this because I don't have enough time to look into this anymore...

mshibuya:
The issue was finally fixed by https://github.com/carrierwaveuploader/carrierwave/commit/556a370c7e5aa2a10b4e3258cd16016452ebb302.
The approach was slightly different, but thank you for giving this idea!

PR #2674, opened 2023-07-14
Fix a `NoMethodError` when a record is rollbacked

y-yagi:
Since v3.0.0, Carrierwave raises `NoMethodError` when a record is rollbacked if uploaders methods don't call at all.

```
NoMethodError:
  private method `select' called for nil:NilClass

    current_paths = (@removed_uploaders + @uploaders.select(&:staged)).map(&:path)
```

This attempts to fix that error and makes work correctly.

### DIFF
--- a/lib/carrierwave/mounter.rb
+++ b/lib/carrierwave/mounter.rb
@@ -202,7 +202,7 @@ module CarrierWave
     end
 
     def remove_added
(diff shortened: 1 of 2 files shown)


mshibuya:
Great, thank you for the fix!

PR #2684, opened 2023-08-17
delete duplicate tests

rajyan:
I was reading carrierwave codebase and noticed that this test is a complete duplicate of 
https://github.com/carrierwaveuploader/carrierwave/blob/d339c275f49de70547c2e5e6145e72fdfdc6379a/spec/orm/activerecord_spec.rb#L879-L887

confirmed that it was added in this commit https://github.com/carrierwaveuploader/carrierwave/commit/8d98ecd19c043b32c97f7745790def0051de94f8 and seems safe to delete


mshibuya:
You're right, thank you!

Issue #2689, opened 2023-08-21
Can't get model.id in store_dir


ogawa5773:
## My Trouble

I'm trying to update the version from 2.x to 3.x.

While checking the operation associated with the version upgrade, we observed a phenomenon that model.id could not be obtained in store_dir and the path was different when uploading images and when referencing them.

### example
```
class BaseImageUploader < CarrierWave::Uploader::Base
  include CarrierWave::MiniMagick

  def store_dir
    "uploads/#{model.class.name.downcase}/#{mounted_as}/#{model.id}"
  end
...
```
If implement the above, the path difference will be like this
- Save like this `uploads/user/thumbnail`
- When referring to `uploads/user/thumbnail/111`

## Question

While debugging, I confirmed that uploading and referencing can be done without problems by deleting the following line. 
https://github.com/carrierwaveuploader/carrierwave/blob/f31546ac41c6a11480425b1528e6834581da2726/lib/carrierwave/orm/activerecord.rb#L46

I read the code there, but I can't understand the relationship that model.id can't be obtained. Can you give me a brief explanation, please? Also is it deprecated to use model.id in store_dir?

## Memo

It has been confirmed that version 2 series (implementation below) is working without problems.
https://github.com/carrierwaveuploader/carrierwave/blob/2f91bee6487d8e5d8bd2c1a88dd25269a2c1e4d0/lib/carrierwave/orm/activerecord.rb#L99

Thanks.

rajyan:
We are using carrierwave 3.x with the same `store_dir` configuration, but it's working fine for us.
I think it's related to the `dup` call in your codebase? (which might be a bug anyway)

rajyan:
@ogawa5773 
I think I could reproduce something similar to what you have reported.
https://github.com/carrierwaveuploader/carrierwave/pull/2690
Does this test express your situation?

ogawa5773:
@rajyan 
Thank you for confirming, that is exactly the phenomenon.
And as far as I verified on my local PC, I feel that the dup method is called internally when ApplicationRecord's update is executed.


For example, if create a User model like the one below and update the name property, the debug log described in initialize_dup will be output.
```
class User < ApplicationRecord

  def initialize_dup(other)
    puts '////////////'
    puts 'initialize_dup called'
    super
  end  
end
```

```
CREATE TABLE users (
  id int(11) NOT NULL auto_increment,
  name varchar(255),
  PRIMARY KEY  (id)
);
```

[...]

PR #2690, opened 2023-08-22
Fix mounter cache duplication on dup

rajyan:
Added a failing test for https://github.com/carrierwaveuploader/carrierwave/issues/2689

duplication should not affect the duplicated objects path.

*Updated*

And now fixes https://github.com/carrierwaveuploader/carrierwave/issues/2689

### DIFF
--- a/lib/carrierwave/orm/activerecord.rb
+++ b/lib/carrierwave/orm/activerecord.rb
@@ -43,6 +43,7 @@ module CarrierWave
         # Reset cached mounter on record dup
         def initialize_dup(other)
           old_uploaders = _mounter(:"#{column}").uploaders
+          @_mounters =  @_mounters.dup
           @_mounters[:"#{column}"] = nil
           super
           # The attribute needs to be cleared to prevent it from picked up as identifier
--- a/lib/carrierwave/orm/activerecord.rb
+++ b/lib/carrierwave/orm/activerecord.rb
@@ -43,7 +43,6 @@ module CarrierWave
         # Reset cached mounter on record dup
         def initialize_dup(other)
           old_uploaders = _mounter(:"#{column}").uploaders
-          @_mounters =  @_mounters.dup
           @_mounters[:"#{column}"] = nil
           super
           # The attribute needs to be cleared to prevent it from picked up as identifier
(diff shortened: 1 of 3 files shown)


spec/orm/activerecord_spec.rb
```diff
@@ -1991,6 +1991,28 @@ def reload(*)
...
+      it "wont affect the duplicated object's saved path" do
+        @event.image = stub_file('test.jpeg')
+        expect(@event.save).to be_truthy
+        expect(@event.image.current_path).to eq public_path("uploads/event/image/#{@event.id}/test.jpeg")
+        new_event = @event.dup
+
+        expect(new_event).not_to be @event
+        expect(new_event.image.model).to be new_event
+        expect(@event.save).to be_truthy
+        expect(@event.image.current_path).to eq public_path("uploads/event/image/#{@event.id}/test.jpeg")
```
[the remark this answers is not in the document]
rajyan: Also, `expect(File.exist?(public_path("uploads/event/image/test.jpeg"))).to be_truthty` passes, so the files is saved to this path. After reloading the model,`current_path` would be changed to `"uploads/event/image/#{@event.id}/test.jpeg"`, but no file is saved in that path. This causes the issue explained in https://github.com/carrierwaveuploader/carrierwave/issues/2689

rajyan:
@mshibuya 
Hope that I've understood the code enough.

lib/carrierwave/orm/activerecord.rb
```diff
@@ -43,6 +43,7 @@ def reload(*)
         # Reset cached mounter on record dup
         def initialize_dup(other)
           old_uploaders = _mounter(:"#{column}").uploaders
+          @_mounters =  @_mounters.dup
```
mshibuya, line 46: Thank you for the PR! Your fix works, but let me propose a slight modification...

As was discussed in https://github.com/carrierwaveuploader/carrierwave/issues/1962#issuecomment-1370308875, this will be executed for every `mount_uploader` / `mount_uploaders`, not just once for a ActiveRecord instance. If we move this to `CarrierWave::Mount::Extension` like 

```diff
--- a/lib/carrierwave/mount.rb
+++ b/lib/carrierwave/mount.rb
@@ -428,6 +428,11 @@ module CarrierWave

     private

+      def initialize_dup(other)
+        @_mounters =  @_mounters.dup
+        super
+      end
+
       def _mounter(column)
         # We cannot memoize in frozen objects :(
         return Mounter.build(self, column) if frozen?
```

I think we can resolve that mismatch. What do you think?

rajyan: Thanks for your suggestion! This looks like a better solution.

mshibuya:
Thank you so much 👍

PR #2691, opened 2023-08-23
fix: call #dup would raise MissingAttributeError while putting :mount_on in options

marsz:
If use :mount_on option and call #dup, it would raise `ActiveModel::MissingAttributeError`

```ruby
class Attachment < ::ApplicationRecord
  mount_uploader(:file, FileUploader, mount_on: :stored_filename)
end

attachment.dup
```

Got error:

```
ActiveModel::MissingAttributeError:
       can't write unknown attribute `file`
```


mshibuya:
Looks good, thank you!

Issue #2695, opened 2023-08-30
Carrierwave download of external url


janz93:
# Problem
Hello everyone 👋 

I am facing errors from Webmock and VCR where specs fail with an unhandled web request error. e.g. `UnhandledHTTPRequestError` or `NetConnectNotAllowedError`. 
During my investigation I noticed that those unhandled web requests are caused by the [`download!`](https://github.com/carrierwaveuploader/carrierwave/blob/8815592942f6c768d432dc6398a441f0c4f6118b/lib/carrierwave/mounter.rb#L134) method. This method gets called if an `external_url` is present. Which is here the case.

# Question
My goal is to have an offline test suite, therefore, I was a bit surprised about this behavior, since I set `enable_processing` to `false`.
I solved the offline approach by mocking all requests:
1. Prevent IP resolving -> Overwrite the `SsrfFilter::DEFAULT_RESOLVER` https://github.com/arkadiyt/ssrf_filter/issues/33#issuecomment-1488800717
2. Mock the external_url URI request -> webmock/vcr

Currently, I am unsure if this is the best approach or if carrierwave should handle this automatically depending on the `enable_processing` setting. Especially since setting `enable_processing` to `false` causes an early exit anyway.

rajyan:
I'm not sure with the best practices of testing carrierwave too, but I'm currently
disabling ssrf_filter by configuring `skip_ssrf_protection?` with a custom downloader only for test
https://github.com/carrierwaveuploader/carrierwave/blob/8815592942f6c768d432dc6398a441f0c4f6118b/lib/carrierwave/downloader/base.rb#L96 

and stubbing the external_url

Maybe some kind of configuration like
https://github.com/carrierwaveuploader/carrierwave/pull/2575/files seems reasonable?

Do you have any opinions? @mshibuya 

These are related issues

https://github.com/arkadiyt/ssrf_filter/issues/33
https://github.com/arkadiyt/ssrf_filter/issues/59
https://github.com/carrierwaveuploader/carrierwave/issues/2573

Issue #2573, opened 2021-06-04
Is CarrierWave::Downloader::Base#skip_ssrf_protection official api?


tashirosota:
Is this usage safe?

```
module CarrierWave
  module Downloader
    class Base
      def skip_ssrf_protection?(_uri)
        true
      end 
    end
  end
end
```

tashirosota:
I want to avoid OpenClass. Can I add a config option?
https://github.com/carrierwaveuploader/carrierwave/pull/2575

PR #2696, opened 2023-09-05
add skip_ssrf_protection config

rajyan:
closes https://github.com/carrierwaveuploader/carrierwave/issues/2695

Looking at these issues, a lot of people are looking for a way to disable ssrf filter in testing.
https://github.com/carrierwaveuploader/carrierwave/issues/2695
https://github.com/arkadiyt/ssrf_filter/issues/33
https://github.com/arkadiyt/ssrf_filter/issues/59
https://github.com/carrierwaveuploader/carrierwave/issues/2573

I believe it might worth making `skip_ssrf_protection` configurable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -659,7 +659,7 @@ end
 ## Testing with CarrierWave
 
 It's a good idea to test your uploaders in isolation. In order to speed up your
-tests, it's recommended to switch off processing in your tests, and to use the
+tests, it's recommended to switch off processing in your tests, disable SSRF protection, and to use the
 file storage. In Rails you could do that by adding an initializer with:
 
 ```ruby
@@ -667,6 +667,7 @@ if Rails.env.test? or Rails.env.cucumber?
   CarrierWave.configure do |config|
     config.storage = :file
     config.enable_processing = false
+    config.skip_ssrf_protection = true
   end
 end
 ```
--- a/README.md
+++ b/README.md
@@ -659,7 +659,11 @@ end
 ## Testing with CarrierWave
 
 It's a good idea to test your uploaders in isolation. In order to speed up your
-tests, it's recommended to switch off processing in your tests, disable SSRF protection, and to use the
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -659,14 +659,15 @@ end
 ## Testing with CarrierWave
 
 It's a good idea to test your uploaders in isolation. In order to speed up your
-tests, it's recommended to switch off processing in your tests, and to use the
+tests, it's recommended to switch off processing in your tests, disable SSRF protection, and to use the
```
mshibuya, line 662: Actually I don't recommend to do this. The users shouldn't do this without knowing what they're doing, unless they will be at risk.
Could you rephrase this part?

rajyan, line 662: > Actually I don't recommend to do this. The users shouldn't do this without knowing what they're doing, unless they will be at risk.

Thank you for your comment. Do you mean there is risk of SSRF attack also during testing?

```suggestion
tests, it's recommended to switch off processing in your tests, and to use the file storage.
Also, you can disable SSRF protection at your own risk using the `skip_ssrf_protection` configuration.

In Rails you could do that by adding an initializer with:

```

What do you think about this explanation?

rajyan: My intention was to use this config during testing and only in testing to avoid outbound requests, because integration tests with webmock fails like explained in https://github.com/arkadiyt/ssrf_filter/issues/33

ssrf_filter performs DNS resolution, which is undetected by UnhandledHTTPRequestError or NetConnectNotAllowedError. Mocking as explained in https://github.com/arkadiyt/ssrf_filter/issues/33#issuecomment-783758611 might make it appear as if the test isn't making outbound requests, but that's not the case in reality.

mshibuya:
Merged in, thank you!

tomasdundacek:
Hi everyone, I have noticed that this feature has been long merged into master branch but not released yet. Do you have any plans on releasing this? 

Thanks a lot for your work :)

PR #2698, opened 2023-09-06
fix: dont retrieve from store until accessing file

Amnesthesia:
Carrierwave seems to try `retrieve_from_store!` as soon as you access `#uploaders`, and if a file doesn't exist on S3 (for whatever reason), it will keep retrying over and over and over, making it virtually unusable because of the amounts of retried requests

It should be possible to access the uploader, and retrieve storage path and identifier without performing network requests to retrieve the file from storage, if retrieving the file isn't needed.

This PR removes `retrieve_from_store!` from the `@uploaders ||=` setup block and instead replaces the `attr_reader :file` with a new method `#file` that will `retrieve_from_store!` when accessed, but only a maximum X times as defined in the configuration for `download_retries` before it just returns `@file` even if it's nil.

### DIFF
--- a/lib/carrierwave/mounter.rb
+++ b/lib/carrierwave/mounter.rb
@@ -66,7 +66,11 @@ module CarrierWave
     def uploaders
       @uploaders ||= read_identifiers.map do |identifier|
         uploader = blank_uploader
-        uploader.retrieve_from_store!(identifier)
+
+        # Assign the identifier to the uploader here,
+        # instead of forcing the uploader to make a round-trip
+        # to the store if we already have the identifier.
+        uploader.identifier = identifier
         uploader
       end
     end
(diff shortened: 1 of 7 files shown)


mshibuya:
Fixed, though not quite where this PR looked.

#retrieve_from_store! never made a request — Storage::Fog#retrieve! only builds a Fog::File from the path. What did make them was everything that ran afterwards: #blank?/#present? asked the storage whether the file was there, and every version's :if/:unless condition was evaluated on retrieval, whether or not the version was ever used.

Both are gone:
- f635d88b9debeda27b25148856ca5e0faa186d17 — #blank? and #present? now only tell whether a file is assigned, with the new #exists? for asking the storage on purpose. The same commit fixes the retrying you saw: directory.files.head was memoized with ||=, so a file which is not there was looked up again on every access. It now remembers a miss.
- 556a370c7e5aa2a10b4e3258cd16016452ebb302 — a version's condition is no longer evaluated until that version is accessed.
Reading a record and using its identifier, path, URL and #present? now makes no requests at all. #exists? makes one, and only one per uploader.

Thanks for the report!

PR #2699, opened 2023-09-07
Test against Ruby 3.2 in CI for 1.x

marcogregorius:
This PR targets the `1.x-stable` branch.

* Adding Ruby 3.2 for tests in CI.
* Use latest `rmagick` for Ruby 3.2 as `rmagick 2.16.0` calls the deprecated `rb_obj_tainted`
  * Also fix the usage for the newer `rmagick` version to use `::Magick::Image.new(width, height) { |img| img.background_color = ...}` pattern.

More context on why 1.x is still being used: https://github.com/carrierwaveuploader/carrierwave/pull/2640#issuecomment-1323532577.

Would also appreciate if the maintainer could release a new gem for 1.x 🙏

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,6 +28,15 @@ jobs:
             gemfile: gemfiles/rails-7-0.gemfile
             experimental: false
           - ruby: '3.1'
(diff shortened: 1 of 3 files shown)


mshibuya:
Thanks, just released 1.3.4.

Issue #2704, opened 2023-09-12
Wrong Content-Type for Adobe Illustrator and Microsoft Office files


shunichi:
Some uploaded files have wrong content-type with carrierwave 3.0.3.

With carrierwave 3.0.3:

* Adobe Illustrator files (`*.ai`) are identified as `application/pdf`.
* Some Excel files (`*.xlsx`) are identified as `application/zip`. (Some Excel files are identified correctly) 

With carrierwave 2.2.4:

* Adobe Illustrator files (`*.ai`) are identified as `application/illustrator`.
* Excel files (`*.xlsx`) are identified as `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`.

I know `*.ai` is a PDF with specific metadata and `*.xlsx` is a ZIP.
But `*.ai` should be downloaded as Adobe Illustrator file.

I think this commit cause the issue.
https://github.com/carrierwaveuploader/carrierwave/commit/a2ca59cbe67046ba7818c64849a9a4ffa90306db

It may be fixed if replcacing this line 
https://github.com/carrierwaveuploader/carrierwave/blob/8815592942f6c768d432dc6398a441f0c4f6118b/lib/carrierwave/sanitized_file.rb#L324
with:
```ruby
Marcel::MimeType.for io, name: original_filename, declared_type: declared_content_type
```

ActiveStorage identified files in the same way.
https://github.com/rails/rails/blob/b5d63b9b16c6f251eaa08f61086f21ff3ef3292f/activestorage/app/models/active_storage/blob.rb#L353-L355

But I don't know it is a correct way to fix the issue. 
It slightly changes the behavior in `guessed_safe_content_type`  case. (in a case where Marcel can't detect the type from the file content and `declared_content_type` is nil)

Juju-62q:
I have similar problem for with `*.csv` files.

With carrierwave 3.0.3:
- Comma-Separated Values files (`*.csv`) are identified as `text/plain`.

With carrierwave 2.2.4:
- Comma-Separated Values files (`*.csv`) are identified as `text/csv`.

With https://github.com/carrierwaveuploader/carrierwave/pull/2705 branch, `*.csv` files are now identified as `text/csv`.

rajyan:
@mshibuya 
Do you have some time looking into this issue?
This might be a blocking issue for some people upgrading to 3.x.

PR #2705, opened 2023-09-12
Improve content type detection for .dotx and .ai

dzhikvas:
To correctly detect content type of file in some cases file extension should be used in addition to magic detection. Such cases include custom extensions with .zip contents, .dotx / .docx files which have same magic signature.

This should also resolve issues described in #2704


mshibuya:
Amazing work, thank you!

shimadama:
@mshibuya 
Sorry if the mentions are incorrect.

I am about to do a major update to v3.0.5.
The commit here is not included and the following occurs.
> Some Excel files (*.xlsx) are identified as application/zip. (Some Excel files are identified correctly)
https://github.com/carrierwaveuploader/carrierwave/issues/2704

Do you plan to release this one?

I referred to the following
https://github.com/carrierwaveuploader/carrierwave/commits/v3.0.5/

Issue #2700, opened 2023-09-07
Upload fails when calling ActiveRecord model 'dup' with after_create or after_update.


macera:
Upload fails when calling ActiveRecord model 'dup' with after_create or after_update.
since version 3.0.0rc

e.g.
[counter_culture](https://github.com/magnusvk/counter_culture)
calls dup on after_create, after_update.

rajyan:
See https://github.com/carrierwaveuploader/carrierwave/issues/2689

I think it's fixed in the latest release (3.0.3).

rajyan:
Sorry, it wasn't released yet...

macera:
Tnanks for letting me know [#2689](https://github.com/carrierwaveuploader/carrierwave/issues/2689) !
It is the same problem.
But I checked on the master branch and the issue was not fixed...

After calling ActiveRecord model 'dup' with `after_create` or `after_update`, 
`@cache_id=nil` happens with `after_save`.

carrierwave/lib/carrierwave/uploader/store.rb
```
def store!(new_file=nil)
  cache!(new_file) if new_file && !cached?
  puts @cache_id #=> nil
  if !cache_only && @file && @cache_id
    with_callbacks(:store, new_file) do
      new_file = storage.store!(@file)
      ...
```
Therefore, it cannot be uploaded.

rajyan:
> But I checked on the master branch and it isn't fixed...

I confirmed the behavior and you are right. https://github.com/carrierwaveuploader/carrierwave/issues/2689 fixed some parts of `dup`, but model with counter_culture still doesn't work (especially on first upload).

rajyan:
> But I checked on the master branch and it isn't fixed...

You're right. I've confirmed that using `counter_cluture` with carrierwave still doesn't work event after fixing `dup` bug in 
https://github.com/carrierwaveuploader/carrierwave/pull/2690 .

The file key is persisted to the DB, but is not stored to the storage.

rajyan:
@macera 

 Could you please try whether [this](https://github.com/carrierwaveuploader/carrierwave/issues/2700) patch fixes your issue?

macera:
@rajyan 
Ohhhh...! I've confirmed that the problem has been fixed !

macera:
@mshibuya 
Do you have a plan for the next release that includes this fix?
I'd appreciate it if you could let me know when.

mshibuya:
Just released 3.0.4 now! 🚀

PR #2706, opened 2023-09-12
Fix dup mounter cache behavior

rajyan:
dup should not prevent uploading

fixes https://github.com/carrierwaveuploader/carrierwave/issues/2700

### DIFF
--- a/spec/orm/activerecord_spec.rb
+++ b/spec/orm/activerecord_spec.rb
@@ -1919,6 +1919,10 @@ describe CarrierWave::ActiveRecord do
       Event.mount_uploader(:image, @uploader)
     end
 
+    after do
+      FileUtils.rm_rf(public_path("uploads"))
+    end
+
     it "caches the existing file into the new model" do
       @event.image = stub_file('test.jpeg')
       @event.save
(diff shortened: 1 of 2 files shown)


rajyan:
I'm very confused with the results, but these test fails inconsistently...

rajyan:
I'm not sure yet with
* If this test is expressing the problem in the issue
* Why the tests are failing randomly

Also, testing locally with counter_culture, I still could not upload an image.

rajyan:
> I'm very confused with the results, but these test fails inconsistently...

This was simply cause by test fixture setup...
The first commit passes, because 'test.jpg' is uploaded by another test and the uploaded files isn't cleaned,

I would fix up them.

rajyan:
All problems are resolved.
I believe this pull request now fixes https://github.com/carrierwaveuploader/carrierwave/issues/2700

rajyan:
@mshibuya 

The last commit in
https://github.com/carrierwaveuploader/carrierwave/pull/2690
broke the `@_mounters` cache, because the `dup` is called after `nil` is set.
Sorry that I couldn't catch it.

spec/orm/activerecord_spec.rb
```diff
@@ -1956,6 +1960,15 @@ def reload(*)
       expect { @event.dup }.not_to change { @event[:image] }
     end
 
+    it "can upload a file" do
```
mshibuya, line 1963:
```suggestion
    it "allows the original object to store a file" do
```

Can the title be a bit more descriptive like this?

mshibuya:
Looks great except for the tiny issue I commented!
This is my mistake as I didn't understand the invocation order of `CarrierWave::Mount::Extension#initialize_dup` and `CarrierWave::ActiveRecord` 's `#initialize_dup` correctly... 😓

spec/orm/activerecord_spec.rb
rajyan: Thanks for your suggestion!

mshibuya:
Thank you so much!

Issue #2702, opened 2023-09-11
ActiveRecord model keeps dirty after update


ssugiyama:
```ruby
class FooUploader < CarrierWave::Uploader::Base
  def filename
    'a.png'
  end
end
```

```ruby
class Foo < ApplicationRecord
  mount_uploader :image, FooUploader
end
```

```ruby
foo = Foo.create!(...)
foo.update!(image: File.open(...))
puts foo.changed? # expecting 'false' but we get 'true'
```

this problem occurs since 3.0.0.rc to 3.0.3
  only if filename method is defined in uploader
  and saved outside transactions (if saved inside transactions, we get 'false')

Are there any workarounds or required settings?

rajyan:
Is this issue related to your problem?
https://github.com/carrierwaveuploader/carrierwave/issues/2682

sdrioux:
Yes, its likely a side effect - but this is still technically a bug that we are running into as well.

ssugiyama:
> Is this issue related to your problem?

Maybe, not. We try changing filename in every save, but the problem is reproduced.

```ruby
class FooUploader < CarrierWave::Uploader::Base
  def filename
    (Time.zone.now.to_f * 1000).to_s + '.png'
  end
end
```

Here is additional information.

```ruby
foo = Foo.create!(...)
foo.update!(image: File.open('newfile.png'))
puts foo.changes
=> {"image"=>["1694578531581.793.png", "newfile.png"]}
```

rajyan:
Thank you for the detailed information!
I think https://github.com/carrierwaveuploader/carrierwave/pull/2707 fixes this bug.

PR #2707, opened 2023-09-13
Fix dirty handling on update

rajyan:
fixes https://github.com/carrierwaveuploader/carrierwave/issues/2702

`mark_remove_#{column}_false` is called after commit
https://github.com/rajyan/carrierwave/blob/e53c9c33674289ffabd8aa6ab3160e2ce60872dd/lib/carrierwave/mount.rb#L387-L389

which was the cause of this issue.
Changed to write a temporary identifier only if remove is changed to a truthy value.

### DIFF
--- a/lib/carrierwave/mounter.rb
+++ b/lib/carrierwave/mounter.rb
@@ -168,7 +168,7 @@ module CarrierWave
 
     def remove=(value)
       @remove = value
-      write_temporary_identifier if remove?
+      write_temporary_identifier
     end
 
     def remove?
--- a/lib/carrierwave/mounter.rb
+++ b/lib/carrierwave/mounter.rb
@@ -168,7 +168,7 @@ module CarrierWave
 
     def remove=(value)
       @remove = value
-      write_temporary_identifier
+      write_temporary_identifier if remove?
     end
 
     def remove?
--- a/spec/orm/activerecord_spec.rb
+++ b/spec/orm/activerecord_spec.rb
@@ -635,18 +635,17 @@ describe CarrierWave::ActiveRecord do
       end
     end
 
-    describe 'with overriddent filename' do
+    describe 'with overridden filename' do
+      before do
+        @uploader.class_eval do
+          def filename
+            model.name + File.extname(super)
+          end
+        end
+        allow(@event).to receive(:name).and_return('jonas')
+      end
 
       describe '#save' do
-
-        before do
-          @uploader.class_eval do
-            def filename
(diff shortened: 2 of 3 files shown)


spec/orm/activerecord_spec.rb
```diff
@@ -664,6 +663,33 @@ def filename
...
+
+        it "shouldn't be generated after second save" do
+          @event.image = stub_file('old.jpeg')
+          @event.save!
+          @event.image = stub_file('new.jpeg')
+          @event.save!
+
+          expect(@event.changes).to be_blank
+          expect(@event).not_to be_changed
+        end
```
rajyan, line 691: The third test is the issue described in #2702 , Not sure yet whether we can make the second one pass.

lib/carrierwave/mounter.rb
```diff
@@ -168,7 +168,7 @@ def blank?
 
     def remove=(value)
       @remove = value
-      write_temporary_identifier
+      write_temporary_identifier if remove?
```
mshibuya, line 171: Doing so is not ideal, as it will prevent users from cancelling the removal by assigning false. (I guess it's better to have a spec for that...)

```ruby
image.remove = true
image.remove = false # this will not call write_temporary_identifier and the attribute change won't be reset
```

I think I found a better solution. Clearing `@original_filename` after store will make `#write_temporary_identifier` harmless, [because `#temporary_identifier` will pick up `@identifier` itself](https://github.com/carrierwaveuploader/carrierwave/blob/96398d9376ef13f47ba4c456b5fd1128e6fcbea8/lib/carrierwave/uploader/proxy.rb#L46). Could you try that approach?

```diff
diff --git a/lib/carrierwave/uploader/store.rb b/lib/carrierwave/uploader/store.rb
index 16a09ca..55d0c06 100644
--- a/lib/carrierwave/uploader/store.rb
+++ b/lib/carrierwave/uploader/store.rb
@@ -87,7 +87,7 @@ module CarrierWave
             end
             @file = new_file
             @identifier = storage.identifier
-            @cache_id = @deduplication_index = nil
+            @original_filename = @cache_id = @deduplication_index = nil
             @staged = false
           end
         end
```

rajyan: Thank you for your review.

I wasn't aware of
> from cancelling the removal by assigning false

this case.

Added some test cases related to remove image cancellation.

mshibuya:
Great job, thank you always 👍

Issue #2708, opened 2023-10-11
The code depends on `@original_filename` doesn't work since v3.0.4


y-yagi:
The `uploader.rb` generated by generators has an example of how to override the filename. The code is the following.

https://github.com/carrierwaveuploader/carrierwave/blob/0c50a8bc7d1c39fe15be056fb22c7b110b9815ae/lib/generators/templates/uploader.rb.erb#L42-L46

But, I think this code doesn't work since v3.0.4 because `@original_filename` is set to nil after `store` by #2707.

So that example always returns `nil` now.  The reproduction script is the following.

```ruby 
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "activerecord"
  gem "sqlite3"
  # gem "carrierwave", "3.0.3"
  gem "carrierwave", "3.0.4"
end

require "active_record"
require "carrierwave"
require "carrierwave/orm/activerecord"
require "minitest/autorun"
require "logger"

ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
ActiveRecord::Base.logger = Logger.new(File::NULL)

ActiveRecord::Schema.define do
  create_table :users, force: true do |t|
    t.string :name
    t.string :avatar
  end
end

class AvatarUploader < CarrierWave::Uploader::Base
  storage :file

  def filename
    "sanitized_filename.txt" if original_filename
  end
end

class User < ActiveRecord::Base
  mount_uploader :avatar, AvatarUploader
end

class BugTest < Minitest::Test
  def test_filename
    u = User.new
    File.open("test.txt") { |f| u.avatar = f }
    u.save!
    assert_equal "sanitized_filename.txt", u.avatar.filename
  end
end
```

```bash
$ ruby bug_report_carrierwave.rb
# Running:

F

Finished in 0.008341s, 119.8938 runs/s, 119.8938 assertions/s.

  1) Failure:
BugTest#test_filename [bug_report_carrierwave.rb:48]:
Expected: "sanitized_filename.txt"
  Actual: nil

1 runs, 1 assertions, 1 failures, 0 errors, 0 skips
```

Is this an intended change?

RachalCassity:
We are also facing this same issue,

In previous versions, the original filename would be stored when called `uploader.store!(file)`. Going through each line, the 
only change was when store! was called. 

[...]

Issue #1898, opened 2016-03-15
original_filename is always present and it makes the filename change on every update


acaron:
This piece of code doesn't work anymore:

```
def filename
    "#{secure_token}.#{file.extension}" if original_filename.present?
end
```

`original_filename` is always present so `update_attributes` or `save` makes the file change its name every single time.

It was the solution to this issue: http://stackoverflow.com/questions/22748385/carrierwave-filename-keeps-changing-on-update-attributes

abMatGit:
I'm also getting an issue with this. Also when loading a model via `Model.find ...`, the uploader will recompute the `secure_token` because it picks up the `original_filename`.

abMatGit:
It seems as though the solution is the memoize the full filename instead of just the secure token.

So instead of:
```ruby
class PhotoUploader < CarrierWave::Uploader::Base
  def filename
    "#{secure_token}.#{file.extension}" if original_filename.present?
  end

  protected
  def secure_token
    var = :"@#{mounted_as}_secure_token"
    model.instance_variable_get(var) or model.instance_variable_set(var, SecureRandom.uuid)
  end
end
``` 

You have:
```ruby
class PhotoUploader < CarrierWave::Uploader::Base
  def filename
    secure_filename if original_filename.present?
  end

  protected

  def secure_filename
    var = :"@#{mounted_as}_secure_filename"
    sec_filename = "#{SecureRandom.uuid}.#{file.extension}"
    model.instance_variable_get(var) or model.instance_variable_set(var, sec_filename)
  end
end
```

abMatGit:
Just as an update, the problem for me was memoizing the `secure_token`. Make sure that if you're saving the same instance of an uploader multiple times, the `secure_token` is memoized to prevent it being recalculated.

acaron:
Thanks @abMatGit, I will try that!

acaron:
No, it doesn't work! `original_filename` returns something even if it is not the original name (it returns the last `#{secure_token}.#{file.extension}"` set and it is suppose to be nil).

1v:
I have same behavior in pure rspec tests, but in capybara tests it working as expected.

aleks:
Any updates on this? We experience the same issue. The secure_token gets shuffled and the reference to the original file is gone once the record is updated.

tglocicero:
Bump

mshibuya:
[...]

PR #2710, opened 2023-10-22
fix to reassign original_filename on store

rajyan:
Trying to fix https://github.com/carrierwaveuploader/carrierwave/issues/2708

### DIFF
--- a/lib/carrierwave/uploader/store.rb
+++ b/lib/carrierwave/uploader/store.rb
@@ -86,8 +86,8 @@ module CarrierWave
               cache_storage.delete_dir!(cache_path(nil))
             end
             @file = new_file
-            @identifier = storage.identifier
-            @original_filename = @cache_id = @deduplication_index = nil
+            @original_filename = @identifier = storage.identifier
+            @cache_id = @deduplication_index = nil
             @staged = false
           end
         end


rajyan:
Still thinking about what origina_filename is expected to be
https://github.com/carrierwaveuploader/carrierwave/issues/1898

Also,

```
diff --git a/lib/carrierwave/uploader/cache.rb b/lib/carrierwave/uploader/cache.rb
index 28f2f77..91e59c5 100644
--- a/lib/carrierwave/uploader/cache.rb
+++ b/lib/carrierwave/uploader/cache.rb
@@ -132,8 +132,8 @@ module CarrierWave
 
         @identifier = nil
         @staged = true
+        self.original_filename = new_file.original_filename
         @filename = new_file.filename
-        self.original_filename = new_file.filename

```

this might be needed?

mshibuya:
Simply put, `@original_filename` is the filename used for storing cache files. I admit that it is named misleadingly, but it's not easy to change it because it was named like that in the first place...

That's why [it's safeguarded not to contain unsafe characters](https://github.com/carrierwaveuploader/carrierwave/blob/0c50a8bc7d1c39fe15be056fb22c7b110b9815ae/lib/carrierwave/uploader/cache.rb#L209). Cache files need to be stored in the filesystem safely. So you just can't use `SanitizedFile#original_filename` that will return unsanitized *original* filename.

rajyan:
Thanks, I think I've might have started to understand some part of original_filename from your explanation.
Before `cache!` in like `RemoteFile#original_filename` or `SantizedFile#original_filename` it stands for the *original* filename.
After cache it's just a cache name.

rajyan:
reference https://github.com/carrierwaveuploader/carrierwave/commit/a2e8a58eac3fb1d07c9487f086c73d185b884a82

rajyan:
@mshibuya 

Digging the histories I'm now confident enough this change should work.

https://github.com/carrierwaveuploader/carrierwave/commit/bea5f998209a4ce110cb2caa08f004f1842420d3

`original_filename` in context of cache, is a identifier for the cache file.
`identifier` is a identifier to retrieve from store.

After store! these can be same. I think assigning `nil` might be more a *correct* value for the current meaning of `original_filename`, but this change should work, because whether the file is `cached?` is only handled by `@cache_id`

https://github.com/carrierwaveuploader/carrierwave/blob/b1dbd7c82f7c1740f18e2bf964dafb9462db1c23/lib/carrierwave/uploader/cache.rb#L78-L80

in all places.

ryan-mcneil:
@mshibuya Hi - I was curious if you intend on approving/merging this "fix", or not as it was not the intended behavior. I'm just wondering if I need to adapt my code to this new behavior (currently pinned to 3.0.3 despite CVE), or if this update will be available soon. Thank you!

mshibuya:
@ryan-mcneil  Thanks for the comment. Then could you answer this question?
https://github.com/carrierwaveuploader/carrierwave/issues/2708#issuecomment-1823931120
It is unlikely to merge this PR, as setting what is not the original filename to `@original_filename` is just misleading. But if there's a very good reason to do so, the situation can change.

ryan-mcneil:
> Then could you answer this question? 

@mshibuya I suppose I don't have a better reason than what has been presented by some of the other engineers on these threads. I'll dig in further while I try to adapt to the new version, and respond again if I come up with something more convincing.

I will side with the sentiment argued by others though. Whether or not the behavior was intended/per the specification, the update did in fact introduce breaking changes for some (many?) users facing a CVE, and likely deserved more than a patch version. Perhaps a 3.0.6 could satisfy these requests, and a 4.X could re-introduce these changes to better align with your intended use case?

Either way, thanks for any consideration, and thanks for doing what you're doing!

mshibuya:
@ryan-mcneil You should have told me that your project is open source. I've opened a draft PR which fixes the specs, at least locally. Please take a look.
https://github.com/department-of-veterans-affairs/vets-api/pull/15118

mshibuya:
Closing as #2718 will alleviate this.

PR #2718, opened 2024-01-13
Show warning if #filename is safeguarded as in the pre-3.x style

### LINKED ISSUE #2708: The code depends on `@original_filename` doesn't work since v3.0.4

mshibuya:
Context: https://github.com/carrierwaveuploader/carrierwave/issues/2708#issuecomment-1890328537

I'm not totally sure that this is side-effect-free (there can be a legitimate case for #filename not to return value?) , but at least it can be helpful to some people.
I want to hear your opinion.
@y-yagi @RachalCassity @rajyan @ryan-mcneil

Closes #2708, Refs. #2659

### DIFF
--- a/lib/carrierwave/uploader/store.rb
+++ b/lib/carrierwave/uploader/store.rb
@@ -14,6 +14,20 @@ module CarrierWave
             @file, @filename, @cache_id, @identifier, @deduplication_index = nil
           end
         }
+
+        after :store, :show_warning_when_filename_is_unavailable
+
+        class_attribute :filename_safeguard_checked
+      end
+
+      module ClassMethods
+      private
+
+        def inherited(subclass)
+          # To perform the filename safeguard check once per a class
+          self.filename_safeguard_checked = false
+          super
+        end
       end
 
       ##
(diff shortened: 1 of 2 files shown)


rajyan:
What I'm not sure is, whether there are other usage for `original_filename`. The change itself seems like it could be beneficial for the 3.x upgrade.

PR #2728, opened 2024-02-29
Use Marcel::Magic.new(content_type).extensions instead of Marcel::TYPES[content_type]

schinery:
This [commit](https://github.com/rails/marcel/commit/e413daeffaed05d1e34b31e0ff1de95182d2d418) of the Marcel gem, which has just shipped in version 1.0.3, breaks Carrierwave as the `Marcel::TYPES` constant used [here](https://github.com/carrierwaveuploader/carrierwave/blob/ed8799191824a4d2762eb1028c01220102699377/lib/carrierwave/downloader/remote_file.rb#L36) has been removed.

This PR is to switch to using `Marcel::Magic.new(content_type).extensions` to return the extensions and not use constants.

### DIFF
--- a/carrierwave.gemspec
+++ b/carrierwave.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
   s.add_dependency "activesupport", ">= 6.0.0"
   s.add_dependency "activemodel", ">= 6.0.0"
   s.add_dependency "image_processing", "~> 1.1"
-  s.add_dependency "marcel", "~> 1.0.3"
+  s.add_dependency "marcel", "~> 1.0.0"
   s.add_dependency "addressable", "~> 2.6"
   s.add_dependency "ssrf_filter", "~> 1.0"
   s.add_development_dependency "csv", "~> 3.0"
--- a/carrierwave.gemspec
+++ b/carrierwave.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
   s.add_dependency "activesupport", ">= 6.0.0"
   s.add_dependency "activemodel", ">= 6.0.0"
   s.add_dependency "image_processing", "~> 1.1"
-  s.add_dependency "marcel", "~> 1.0.0"
+  s.add_dependency "marcel", "~> 1.0.3"
   s.add_dependency "addressable", "~> 2.6"
   s.add_dependency "ssrf_filter", "~> 1.0"
   s.add_development_dependency "csv", "~> 3.0"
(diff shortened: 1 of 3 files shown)


schinery:
Not sure why, but this PR is getting the following failure in the ruby-head tests for `./spec/processing/rmagick_spec.rb:248`.

```
expected NoMethodError with message matching /private method `foo=' called/, got #<NoMethodError: private method 'foo=' called for an instance of Magick::Image::Info> with backtrace:
```

Doesn't seem to be related to these changes, but flagging anyway.

BrianHawley:
The test failure in ruby-head is because of https://bugs.ruby-lang.org/issues/16495

In spec/processing/rmagick_spec.rb line 258, change:
```ruby
/private method `foo=' called/
```
to:
```ruby
/private method .foo=. called/
```

That will handle the quoting change.

bvogel:
I would vote not not move from one **private** constant to another

lib/carrierwave/downloader/remote_file.rb
```diff
@@ -33,9 +33,9 @@ def headers
 
       def original_filename
         filename = filename_from_header || filename_from_uri
-        mime_type = Marcel::TYPES[content_type]
-        unless File.extname(filename).present? || mime_type.blank?
-          extension = mime_type[0].first
+        extensions = Marcel::TYPE_EXTS[content_type]
```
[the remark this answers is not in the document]
schinery: Yeah, that is much nicer 😄

carrierwave.gemspec
```diff
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
   s.add_dependency "activesupport", ">= 6.0.0"
   s.add_dependency "activemodel", ">= 6.0.0"
   s.add_dependency "image_processing", "~> 1.1"
-  s.add_dependency "marcel", "~> 1.0.0"
+  s.add_dependency "marcel", "~> 1.0.3"
```
BrianHawley, line 26: With the new code change, you don't need to make this version change, because the code is supported as far back as version 1.0.0.

schinery: Reverted this change 👍🏻

BrianHawley (approved):

bvogel:
Nice work everyone!

bvogel (approved):

mshibuya:
Thank you for the fix!

bvogel:
@mshibuya any chance for a new release soon?

PR #2733, opened 2024-03-21
Add actual IM error to output when failing to process

### LINKED ISSUE #2732: When failing to process with ImageMagick, could we include the actual processing ImageMagick error message?
I had upgraded my Alpine image and didn't realize [Alpine no longer includes all the different processing/decode delegates by default](https://gitlab.alpinelinux.org/alpine/aports/-/issues/15096).

It could have saved me some hours to have the actual error message from ImageMagick in this case. Maybe this could help someone else.

gblair:
Fixes #2732 2732


mshibuya:
The error message was intentionally removed in #2499.
There can be another way to let developers know what's happening, but at least showing end-users those validation error messages doesn't help them.

PR #2734, opened 2024-03-27
bump image_processing version to avoid project being stuck on old version

dangerousbeans:
To resolve my app being stuck on 1.1, which is causing problems with 

```
Active Storage's ImageProcessing transformer doesn't support :combine_options, as it always generates a single ImageMagick command.`
```

https://github.com/rails/rails/issues/41322#issuecomment-899719446


mshibuya:
It's unrelated with the carrierwave's dependency. What you need to do is to run `bundle update` with your app.

PR #2744, opened 2024-06-12
CI against Ruby 3.3

y-yagi:
This PR adds Ruby 3.3 to the CI matrix to confirm `carrierwave` working with Ruby 3.3.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -18,10 +18,10 @@ jobs:
           - ruby: '2.6'
             gemfile: gemfiles/rails-6-1.gemfile
             experimental: false
-          - ruby: '3.2'
+          - ruby: '3.3'
```
y-yagi, line 21: It seems that the latest Ruby was specified. So I updated here and L24 too.

mshibuya:
Thank you!

Issue #2722, opened 2024-01-30
Version names starting with underscore and numbers lead to NameError


langalex:
Given an uploader:

```ruby
class MyUploader < CarrierWave::Uploader::Base
  version :_800x600 do
    process resize_to_fill: [800, 600]
  end
end
```

leads to this error:

```
NameError:
  wrong constant name 800x600VersionUploader

# [..]/carrierwave-3.0.5/lib/carrierwave/uploader/versions.rb:23:in `const_set'
# [..]/carrierwave-3.0.5/lib/carrierwave/uploader/versions.rb:23:in `build'
# [..]/carrierwave-3.0.5/lib/carrierwave/uploader/versions.rb:176:in `block in versions'
# [..]/carrierwave-3.0.5/lib/carrierwave/uploader/versions.rb:175:in `each'
# [..]/carrierwave-3.0.5/lib/carrierwave/uploader/versions.rb:175:in `versions'
# [..]/carrierwave-3.0.5/lib/carrierwave/uploader/versions.rb:301:in `active_versions'
# [..]/carrierwave-3.0.5/lib/carrierwave/uploader/versions.rb:309:in `derived_versions'
# [..]/carrierwave-3.0.5/lib/carrierwave/uploader/versions.rb:329:in `cache_versions!'
# [..]/carrierwave-3.0.5/lib/carrierwave/uploader/callbacks.rb:16:in `block in with_callbacks'
# [..]/carrierwave-3.0.5/lib/carrierwave/uploader/callbacks.rb:16:in `each'
# [..]/carrierwave-3.0.5/lib/carrierwave/uploader/callbacks.rb:16:in `with_callbacks'
# [..]/carrierwave-3.0.5/lib/carrierwave/uploader/cache.rb:146:in `cache!'
# [..]/carrierwave-3.0.5/lib/carrierwave/uploader/versions.rb:234:in `cache!'
# [..]/carrierwave-3.0.5/lib/carrierwave/uploader/store.rb:80:in `store!'
```

A possible solution could be to prefix [the generated class name](https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/uploader/versions.rb#L23).

PR #2748, opened 2024-08-07
Change version constant naming pattern

r7kamura:
Because of the restriction that Ruby constant names must always begin with `[A-Z]`, it would be desirable to make sure that this restriction is satisfied when naming dynamic constants.

- Fix https://github.com/carrierwaveuploader/carrierwave/issues/2722

It was implemented well in carrierwave v2, but when `object_id` was removed from the constant name  at https://github.com/carrierwaveuploader/carrierwave/commit/a9de7565eabb4cdca05bb090cdf797eb1720c09c, `Uploader` prefix was also removed at the same time for some reason.

### DIFF
--- a/lib/carrierwave/uploader/versions.rb
+++ b/lib/carrierwave/uploader/versions.rb
@@ -20,7 +20,7 @@ module CarrierWave
         def build(superclass)
           return @klass if @klass
           @klass = Class.new(superclass)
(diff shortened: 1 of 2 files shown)


mshibuya:
Ah I didn't realize this point. Thank you so much!

PR #2750, opened 2024-08-14
fix URLs

yasudatoshiyuki:
I found the dead links while I was reading the docs around ‘processing’, so I replaced them with new ones.


mshibuya:
Thank you 👍

PR #2752, opened 2024-08-29
Backport security fixes for 1.x-stable

mbie:
There are two security vulnerabilities (https://github.com/carrierwaveuploader/carrierwave/security/advisories/GHSA-vfmv-jfc5-pjjw and https://github.com/carrierwaveuploader/carrierwave/security/advisories/GHSA-gxhx-g4fq-49hj) which were resolved for 2.x and 3.x versions. They have not been addressed in 1.x branch.

I'm aware that using 1.x should not happening anymore, but in our case migration to 3.0 is a painful way which will take time. This backport would allow us to resolve the security vulnerabilities from our systems.

### DIFF
--- a/lib/carrierwave/sanitized_file.rb
+++ b/lib/carrierwave/sanitized_file.rb
@@ -266,7 +266,7 @@ module CarrierWave
     def content_type
       return @content_type if @content_type
       if @file.respond_to?(:content_type) and @file.content_type
-        @content_type = @file.content_type.to_s.chomp
+        @content_type = ::MIME::Types[@file.content_type.to_s.chomp].first.to_s
       elsif path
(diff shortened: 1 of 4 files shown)


mshibuya:
Yeah it'll take a real effort to make it functional again. I want to avoid the maintenance work if possible...

What's the blocker for 3.0 upgrade?

mbie:
In our case it is just matter of huge codebase and a lot of monkey patches. Nothing that we cannot adjust to 3.x but it will take more time. 

We will do it anyway in long term, but we thought of having simple and quick alternative short-term solution. 

If the backport for 1.x is indeed not quick, we will focus on the upgrade to 3.

PR #2755, opened 2024-10-02
CI against Rails 7.2

y-yagi:
This PR adds Rails 7.2 to the CI matrix to confirm `carrierwave` working with Rails 7.2.

Currently, the latest Rails runs against only the latest Ruby and the previous Rails version runs against multiple Rubies. So I updated the matrix to follow it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,8 +8,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ '2.7', '3.0', '3.1', '3.2', '3.3', jruby ]
-        gemfile: [ gemfiles/rails-7-0.gemfile ]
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -8,8 +8,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ '2.7', '3.0', '3.1', '3.2', '3.3', jruby ]
```
y-yagi, line 11: I removed this because `activerecord-jdbc-adapter` doesn't support Rails 7.1 yet 
https://github.com/jruby/activerecord-jdbc-adapter/issues/1146

Instead of this, I added to the matrix for JRuby + gemfiles/rails-7-0.gemfile.

seuros (approved):

mshibuya:
Thank you 🙌

Issue #2759, opened 2024-11-07
Incompatible with URI >=1.0.0


swelther:
Today the standard gem URI was released as 1.0.0, introducing a new default parser `RFC3986_Parser`: https://github.com/ruby/uri/pull/107/files

This results in errors like

```
NameError: uninitialized constant URI::REGEXP
Did you mean?  Regexp
…/ruby/3.1.0/gems/uri-1.0.0/lib/uri/common.rb:53:in `const_missing'
…/ruby/3.1.0/gems/carrierwave-3.0.7/lib/carrierwave/utilities/uri.rb:7:in `<module:Uri>'
```

I think I could prevent this error by setting the default parser back to the old one with `URI.parser = URI::RFC2396_PARSER`. But not sure how carrierwave should behave.

jules-w2:
i confirm, @ollym 's PR fix this issue.
I hope it can be merged quickly

taketo1113:
I encountered the same error with uri gem v1.0.0.

Updating to uri gem v1.0.1 resolved the `uninitialized constant URI::REGEXP` error.
It seems that in uri gem v1.0.1, `URI::REGEXP` fallback to `URI::RFC2396_REGEXP`.

https://github.com/ruby/uri/issues/125#issuecomment-2463863189

PR #2760, opened 2024-11-07
Use RFC2396_REGEXP explicitly

ollym:
Because uri > 1.0 switched to RFC3986 as the default parser

Fixes https://github.com/carrierwaveuploader/carrierwave/issues/2759


ollym:
@mshibuya if you can take a look as the patch changes you made to the original file. Thanks

mshibuya:
Thank you!

PR #2763, opened 2024-11-18
Support AWS S3 FIPS endpoints

matt-domsch-sp:
S3 endpoints even when ENV['AWS_USE_FIPS_ENDPOINT']=='true'. When FIPS is called for, and we are in a region where FIPS endpoints are available, this method should return the FIPS endpoint.

Furthermore, when S3 Transfer Acceleration (S3TA) is requested by configuration, the above endpoint gets overridden to select the S3TA endpoint. However, S3TA is not avaialble in GovCloud, and has no FIPS endpoint equivalents. In this instance, if the region is a GovCloud region, or if FIPS mode is called for, do not override the endpoint to use S3TA.

This is functionally equivalent to an issue submitted to the fog-aws project.  https://github.com/fog/fog-aws/issues/729

### DIFF
--- a/lib/carrierwave/storage/fog.rb
+++ b/lib/carrierwave/storage/fog.rb
@@ -395,7 +395,7 @@ module CarrierWave
                 # GovCloud doesn't support S3 Transfer Acceleration https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/govcloud-s3.html
                 # S3 Transfer Acceleration doesn't support FIPS endpoints.  When both fog_aws_accelerate=true and AWS_USE_FIPS_ENDPOINT=true, don't use Accelerate.
                 if @uploader.fog_aws_accelerate && (AWS_GOVCLOUD_REGIONS.include?(region) || ENV['AWS_USE_FIPS_ENDPOINT'] == 'true')
-                  warn "S3 Transfer Acceleration is not available in GovCloud regions or when AWS_USE_FIPS_ENDPOINT=true. Disabling accelearation."
+                  warn "S3 Transfer Acceleration is not available in GovCloud regions or when AWS_USE_FIPS_ENDPOINT=true. Disabling acceleration."
                   @uploader.fog_aws_accelerate = false
                 end
 
--- a/lib/carrierwave/storage/fog.rb
+++ b/lib/carrierwave/storage/fog.rb
@@ -395,6 +395,7 @@ module CarrierWave
                 # GovCloud doesn't support S3 Transfer Acceleration https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/govcloud-s3.html
                 # S3 Transfer Acceleration doesn't support FIPS endpoints.  When both fog_aws_accelerate=true and AWS_USE_FIPS_ENDPOINT=true, don't use Accelerate.
                 if @uploader.fog_aws_accelerate && (AWS_GOVCLOUD_REGIONS.include?(region) || ENV['AWS_USE_FIPS_ENDPOINT'] == 'true')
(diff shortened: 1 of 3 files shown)


matt-domsch-sp:
The equivalent patch to fog-aws has been merged into that upstream.

lib/carrierwave/storage/fog.rb
```diff
@@ -383,12 +385,19 @@ def public_url
...
+                regional_host = 's3.amazonaws.com' # used for DEFAULT_S3_REGION or no region set
+                if ENV['AWS_USE_FIPS_ENDPOINT'] == 'true' && AWS_FIPS_REGIONS.include?(region)
+                  regional_host = "s3-fips.#{region}.amazonaws.com" # https://aws.amazon.com/compliance/fips/
+                elsif ![DEFAULT_S3_REGION, ''].include?(region)
+                  regional_host = "s3.#{region}.amazonaws.com"
+                end
+
+                # GovCloud doesn't support S3 Transfer Acceleration https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/govcloud-s3.html
+                # S3 Transfer Acceleration doesn't support FIPS endpoints.  When both fog_aws_accelerate=true and AWS_USE_FIPS_ENDPOINT=true, don't use Accelerate.
+                if @uploader.fog_aws_accelerate && (AWS_GOVCLOUD_REGIONS.include?(region) || ENV['AWS_USE_FIPS_ENDPOINT'] == 'true')
```
mshibuya, line 397: We don't need to embed the knowledge about AWS regions into CarrierWave. If an app is using AWS regions that supports S3 Transfer Acceleration it can be set as `config.fog_aws_accelerate = true`, that's it. We should let developers decide.

lib/carrierwave/storage/fog.rb
```diff
@@ -383,12 +385,19 @@ def public_url
...
 
                 region = @uploader.fog_credentials[:region].to_s
-                regional_host = case region
-                                when DEFAULT_S3_REGION, ''
-                                  's3.amazonaws.com'
-                                else
-                                  "s3.#{region}.amazonaws.com"
-                                end
+                regional_host = 's3.amazonaws.com' # used for DEFAULT_S3_REGION or no region set
+                if ENV['AWS_USE_FIPS_ENDPOINT'] == 'true' && AWS_FIPS_REGIONS.include?(region)
```
mshibuya, line 389: Don't use an environment variable here. You should create a config like `fog_aws_fips`, in a same manner with `fog_aws_accelerate`.  You don't need to check AWS region here, just letting developer choose is enough.

mshibuya:
A change needs to have corresponding RSpec test examples, so it doesn't break unintentionally in future.

lib/carrierwave/storage/fog.rb
matt-domsch-sp: ok, I can remove this section.

lib/carrierwave/storage/fog.rb
matt-domsch-sp: This is the same environment variable that the AWS SDK uses to determine if _it_ should choose the FIPS endpoints, hence my use of it here.

mshibuya: CarrierWave is not the AWS SDK. Every software has its own way of configuration, and you need to follow it. Otherwise there'll be a mess.

matt-domsch-sp:
Thank you for the review and feedback.  Code cleaned up, :fog_aws_fips option and specs added.  The added specs fail the Cyclomatic complexity rule. Please advise how you'd like this to be handled.

mshibuya:
OK I can take care of the rest. Thank you!

PR #2767, opened 2024-12-17
Support test matchers with only Vips

gaffneyc:
This adds support for using the test matchers when using Vips and neither MiniMagick or Magick are installed. In our case it was falling back to Magick since MiniMagick wasn't available. I didn't see any tests for the matchers and I'm not sure if there is even a good way to test the switching logic. That said, I've tested these changes against our application and they fixed the issue we were seeing.

The one idiosyncrasy is that there isn't a clean way to get the format like with the Magick libraries. The best Vips can do is to tell which loader was used (e.g. `"pngload"`) which I'm guessing is based on the extension.

### DIFF
--- a/lib/carrierwave/test/matchers.rb
+++ b/lib/carrierwave/test/matchers.rb
@@ -336,6 +336,8 @@ module CarrierWave
         def self.load_image(filename)
           if defined? ::MiniMagick
             MiniMagickWrapper.new(filename)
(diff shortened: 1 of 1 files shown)


w-masahiro-ct (approved):
LGTM

mshibuya:
Thanks!

PR #2773, opened 2025-01-08
Remove use of deprecated mb_chars method

chaadow:
mb_chars has been deprecated. and can be replaced by simply removing the method.

cf : https://github.com/rails/rails/pull/54081


mshibuya:
Makes much sense 👍

PR #2783, opened 2025-03-29
Fix test CI

w-masahiro-ct:
## What's changed

- Fix test CI

*The test for `ruby-head` fail, but since `experimental: true` is set, it seems it's not likely to be a problem.


mshibuya:
You need to describe what problem your change is trying to address. Please focus only on that, and don't put unrelated changes together.

PR #2785, opened 2025-05-06
Add content_type_allowlist example to uploader template

coloradosazulyanaranjados:
When developers add CarrierWave for the first time they usually follow the docs and generate an uploader with the provided generator.  At the moment, that generator does not include a `content_type_allowlist` example, which can leave newcomers vulnerable to content-type spoofing if they overlook the documentation.

This PR adds a commented `content_type_allowlist` example, in the same style as the existing `extension_allowlist` and other examples in the template.


mshibuya:
Thank you!

PR #2793, opened 2025-10-03
Don't call s3 multiple times when the file doesn't exist

nerdrew:
When s3 returns a 404, the file is not cached, so every access will re-attempt to fetch the file. Instead, cache the result of the request, even when it returns a 404.


mshibuya:
The change was incorporated into https://github.com/carrierwaveuploader/carrierwave/commit/f635d88b9debeda27b25148856ca5e0faa186d17. Thanks!

PR #2805, opened 2026-02-11
Ensure ruby 4 compatibility

wynksaiddestroy:
I added ruby 4.0 to the CI test matrix.
I fixed a test that already failed before.
And I needed to remove the pinning of the sqlite3 gem.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ '3.2', '3.3', '3.4' ]
+        ruby: [ '3.2', '3.3', '3.4', '4.0' ]
         gemfile: [ gemfiles/rails-8-0.gemfile ]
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -30,6 +30,9 @@ jobs:
           - ruby: '3.4'
             gemfile: gemfiles/rails-main.gemfile
```
wynksaiddestroy, line 31: Wouldn't be `rails-8-0.gemfile` the better choice here?

.github/workflows/test.yml
```diff
@@ -30,6 +30,9 @@ jobs:
...
             experimental: true
```
wynksaiddestroy, line 32: Should I change this to `false`?

mshibuya:
Closing in favor of https://github.com/carrierwaveuploader/carrierwave/pull/2815.
Thank you for the contribution anyway!

PR #2808, opened 2026-02-25
Fix Firebase Storage URL decoding (%2F)

maiiat:
#### Description:

  #### Problem

  Firebase Storage URLs use `%2F` (encoded slash) as part of object identifiers.
  `process_uri` currently decodes `%2F` → `/`, which breaks these URLs.

  - Before (original URL): `/o/images%2Fsubfolder%2Fphoto.jpg` ✅
  - After `process_uri`: `/o/images/subfolder/photo.jpg` ❌ 400

  #### Solution

  Added `preserve_path_encoding?(host)` to skip path decode/encode for Firebase Storage URLs. The hook can be overridden for custom hosts.

  #### Changes

  - Preserve %2F encoding in path for firebasestorage.googleapis.co
  - Case-insensitive host matching
  - Added tests for Firebase URLs
  - Behavior for non-Firebase URLs remains unchanged

### DIFF
--- a/lib/carrierwave/downloader/base.rb
+++ b/lib/carrierwave/downloader/base.rb
@@ -67,12 +67,41 @@ module CarrierWave
         uri = Addressable::URI.parse(source)
         uri.host = uri.normalized_host
         # Perform decode first, as the path is likely to be already encoded
-        uri.path = encode_path(decode_uri(uri.path)) if uri.path =~ CarrierWave::Utilities::Uri::PATH_UNSAFE
+        # Skip decode/encode for hosts that use encoded slashes as part of object identifiers (e.g. Firebase Storage)
+        if uri.path =~ CarrierWave::Utilities::Uri::PATH_UNSAFE && !preserve_path_encoding?(uri.host)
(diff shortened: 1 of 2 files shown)


maiiat:
@mshibuya @w-masahiro-ct Firebase + CarrierWave = 400 Bad Request. Until now. 😎 Every Firebase user hits this - 3 lines fix it forever 🔥

mshibuya:
Thanks for the PR, but got a different fix: https://github.com/carrierwaveuploader/carrierwave/commit/c2a182ef8c414eb43cca553ec06380075ea18a48

maiiat:
> Thanks for the PR, but got a different fix: [c2a182e](https://github.com/carrierwaveuploader/carrierwave/commit/c2a182ef8c414eb43cca553ec06380075ea18a48)

Thanks for fixing this in master. Do you have an **approximate plan for the next gem release** that will include c2a182e? We use CarrierWave via Gemfile.lock and would prefer to avoid pinning to an unreleased GitHub commit.

mshibuya:
It will be shipped as a part of the next major release. No concrete plan but maybe a few weeks?

PR #2809, opened 2026-03-20
Fix process_uri rewriting literal + in URL paths to %2B (#2800)

### LINKED ISSUE #2800: `CarrierWave::Downloader::Base#process_uri` changes `+` to `%2B` in the path, breaking valid URLs
I am running into an issue with `CarrierWave::Downloader::Base#process_uri` where a valid URL becomes invalid after CarrierWave processes it.

The URL I provide is **not encoded**. It contains a literal `+` in the **path**. That `+` is valid for this server and must remain a literal `+`. However, CarrierWave’s URI processing converts it to `%2B`, and the resulting URL returns `404 Not Found`.

Example:

```
Input URL (works):
https://pokemoncardimages.pokedata.io/images/Red+Flash/016.webp

URL after CarrierWave processing (fails):
https://pokemoncardimages.pokedata.io/images/Red%2BFlash/016.webp
```

Because of this, assigning something like:

```ruby
model.remote_image_url = 'https://pokemoncardimages.pokedata.io/images/Red+Flash/016.webp'
```

ends up raising:

```
could not download file: 404 "Not Found"
```

To work around the problem I currently have a monkey patch for `process_uri` that normalizes using `Addressable::URI` while preserving the `+` in the path, but I would prefer not to patch CarrierWave internals.

Would you consider adjusting `process_uri` so that it does not percent-encode `+` characters in the path (or otherwise preserves the original path when it is already a valid URI for HTTP requests)?

I can provide a minimal reproduction if needed.

Thanks.

RYOTA-KOBA:
## Summary

Fixes #2800

`process_uri` was converting literal `+` signs in URL paths to `%2B`
before issuing the HTTP request, causing 404 errors on servers that use
`+` as a valid path character (e.g. `/my+test.jpg`)

## Root Cause

The old implementation used a decode-then-re-encode strategy:

```ruby
# Before
uri.path = encode_path(decode_uri(uri.path)) if uri.path =~ PATH_UNSAFE
```

`decode_uri` unconditionally decoded all `%XX` sequences, turning `%2B`
into `+`. Then `encode_path` re-encoded everything with `PATH_UNSAFE`,
which treated `+` as unsafe — converting all `+` (both originally
literal and originally `%2B`) to `%2B`. This made it impossible to
distinguish between a literal `+` and an intentionally encoded `%2B`.

## Solution

Replace the decode-then-re-encode strategy with a **single-pass
encoding** that skips existing `%XX` sequences:

```ruby
# After
uri.path = encode_path_preserving_encoding(uri.path, PATH_UNSAFE_WITH_PLUS) if uri.path =~ PATH_UNSAFE_WITH_PLUS
```

- `encode_path_preserving_encoding` uses a regex alternation
  `(%[0-9a-fA-F]{2})|(unsafe_char)` to match the path: existing
  percent-encoded sequences are passed through untouched (`$1`), while
  truly unsafe characters are percent-encoded byte-by-byte (`$2`).
- `PATH_UNSAFE_WITH_PLUS` extends `PATH_UNSAFE` to treat `+` as a safe
  character, per RFC 3986 §3.3 which permits `+` as a literal in paths.

## CI Notes

The failing checks are reproducible on the current base and are not introduced by this PR.

Observed unrelated failures:
- RuboCop: `lib/carrierwave/uploader/configuration.rb` (`Lint/Void`)
- `jruby-head`: experimental/dependency-level instability

### DIFF
--- a/lib/carrierwave/downloader/base.rb
+++ b/lib/carrierwave/downloader/base.rb
@@ -66,8 +66,7 @@ module CarrierWave
       def process_uri(source)
         uri = Addressable::URI.parse(source)
         uri.host = uri.normalized_host
-        # Perform decode first, as the path is likely to be already encoded
-        uri.path = encode_path(decode_uri(uri.path)) if uri.path =~ CarrierWave::Utilities::Uri::PATH_UNSAFE
+        uri.path = encode_path_preserving_encoding(uri.path, PATH_UNSAFE_WITH_PLUS) if uri.path =~ PATH_UNSAFE_WITH_PLUS
         uri.query = encode_non_ascii(uri.query) if uri.query
         uri.fragment = encode_non_ascii(uri.fragment) if uri.fragment
         URI.parse(uri.to_s)
--- a/lib/carrierwave/utilities/uri.rb
+++ b/lib/carrierwave/utilities/uri.rb
@@ -6,6 +6,8 @@ module CarrierWave
       # based on Ruby < 2.0's URI.encode
       PATH_SAFE = URI::RFC2396_REGEXP::PATTERN::UNRESERVED + '\/'
       PATH_UNSAFE = Regexp.new("[^#{PATH_SAFE}]", false)
+      PATH_SAFE_WITH_PLUS = PATH_SAFE + '\+'
+      PATH_UNSAFE_WITH_PLUS = Regexp.new("[^#{PATH_SAFE_WITH_PLUS}]", false)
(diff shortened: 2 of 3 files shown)


Cremesis:
Any update on this PR?

mshibuya:
Thanks for the PR, but got a different fix: https://github.com/carrierwaveuploader/carrierwave/commit/c2a182ef8c414eb43cca553ec06380075ea18a48

PR #2811, opened 2026-03-26
Add fog_acl config to allow disabling or customizing storage ACLs

stanhu:
AWS changed the default S3 bucket settings in April 2023 to disable ACLs, causing AccessControlListNotSupported errors when CarrierWave sends x-amz-acl headers. The new fog_acl option decouples ACL behaviour from fog_public:

- fog_acl = nil (default): existing behaviour, ACL derived from fog_public
- fog_acl = false: omits ACL entirely (no :public passed to fog, no ACL header on copy); suitable for buckets/containers with ACLs disabled
- fog_acl = 'bucket-owner-full-control' (or any string): uses that value as x-amz-acl (AWS) or destination_predefined_acl (Google) on copy; pair with fog_attributes to apply the same header on store operations


stanhu:
https://github.com/carrierwaveuploader/carrierwave/pull/2810 fixes the issue addressed in d74c3b0a5af5070f96e5c7731ef300a54376de2f.

mshibuya:
Thank you!

PR #2813, opened 2026-04-03
Add ssrf_filter_options configuration to pass options to SsrfFilter to use a static IPv4 address

kaibadash:
CarrierWave uses SsrfFilter for SSRF protection when downloading remote files, but does not expose SsrfFilter's options (such as `resolver`, `http_options`, etc.) to users.

This is problematic when SsrfFilter resolves a hostname that returns both IPv4 and IPv6 addresses. SsrfFilter randomly picks one via `.sample`, and if IPv6 is selected but unreachable on the server, the download fails with a network error. Users have no way to control DNS resolution (e.g., to prefer IPv4) without monkey-patching.

https://github.com/arkadiyt/ssrf_filter/blob/b87b175f955743c8abee4b37053d92fe3da512c9/lib/ssrf_filter/ssrf_filter.rb#L159

This commit adds `ssrf_filter_options` configuration that passes through any options to `SsrfFilter.get`, allowing users to customize behavior such as providing a custom resolver.

Example usage:

      config.ssrf_filter_options = {
        resolver: proc { |hostname|
          Resolv.getaddresses(hostname)
            .map { |ip| IPAddr.new(ip) }
            .select(&:ipv4?)
        }
      }
    end

### DIFF
--- a/lib/carrierwave/downloader/base.rb
+++ b/lib/carrierwave/downloader/base.rb
@@ -32,12 +32,13 @@ module CarrierWave
             response = OpenURI.open_uri(process_uri(url.to_s), headers)
           else
             request = nil
+            ssrf_options = @uploader.ssrf_filter_options || {}
             if ::SsrfFilter::VERSION.to_f < 1.1
-              response = SsrfFilter.get(uri, headers: headers) do |req|
+              response = SsrfFilter.get(uri, headers: headers, **ssrf_options) do |req|
                 request = req
               end
             else
-              response = SsrfFilter.get(uri, headers: headers, request_proc: ->(req) { request = req }) do |res|
+              response = SsrfFilter.get(uri, headers: headers, request_proc: ->(req) { request = req }, **ssrf_options) do |res|
                 res.body # ensure to read body
               end
             end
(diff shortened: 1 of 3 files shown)


muk-ai:
We've been hitting this exact issue — when `SsrfFilter` picks an IPv6 address on a host without IPv6 reachability, the download fails with `Network is unreachable`. This started causing a large number of errors over the past week.

We'd love to see this merged. Thank you for this PR!

kaibadash:
@muk-ai If you're in a hurry, try using a monkey patch.
```ruby
module SsrfFilterIpv4Resolver
  def get(uri, **options, &block)
    options[:resolver] ||= proc {|hostname|
      Resolv.getaddresses(hostname).
        map {|ip| IPAddr.new(ip) }.
        select(&:ipv4?)
    }
    super
  end
end
SsrfFilter.singleton_class.prepend(SsrfFilterIpv4Resolver)
```

muk-ai:
> @muk-ai If you're in a hurry, try using a monkey patch.
> 
> ```ruby
> module SsrfFilterIpv4Resolver
>   def get(uri, **options, &block)
>     options[:resolver] ||= proc {|hostname|
>       Resolv.getaddresses(hostname).
>         map {|ip| IPAddr.new(ip) }.
>         select(&:ipv4?)
>     }
>     super
>   end
> end
> SsrfFilter.singleton_class.prepend(SsrfFilterIpv4Resolver)
> ```

@kaibadash 
Thank you for the suggestion! We've already applied a temporary workaround on our side, but I really appreciate you taking the time to share the code example. 🙏

mshibuya:
Thanks!

PR #2814, opened 2026-04-03
Fix Lint/Void warning in storage method

kaibadash:
The `storage` variable in the `when nil` branch was used in void context since the return value comes from `_storage` at the end of the method.


kaibadash:
dup https://github.com/carrierwaveuploader/carrierwave/pull/2810

PR #2815, opened 2026-04-28
CI against Ruby 4.0

y-yagi:
This PR adds Ruby 4.0 to the build matrix to confirm whether CarrierWave works with it.


y-yagi:
CI has broken. I will try to fix it with #2816

mshibuya:
Thank you!

PR #2817, opened 2026-05-11
Raise `ProcessingError` when all frames are filtered out `in manipulate!`

y-yagi:
Previously, when `manipulate!` was called with a block that returned nil for every frame, the resulting empty frames array caused an `ArgumentError` in `frames.append(true)`. This is because RMagick's `ImageList#append` does not accept a bare true value on an empty list.

https://github.com/rmagick/rmagick/blob/9788a6d43b3aebb5a1ad72013f1fe0b3affa300f/ext/RMagick/rmilist.cpp#L143 https://github.com/rmagick/rmagick/blob/9788a6d43b3aebb5a1ad72013f1fe0b3affa300f/ext/RMagick/rmilist.cpp#L877

This made the error hard to diagnose since the root cause (unsupported format or a block that always returns nil) was not surfaced.

It looks like this happens when an image file isn't supported by ImageMagick.(In our case, this happens with HEIC files with unsupported compatible brands)

So this changed to check frames and if it's empty, raises an appropriate error.

### DIFF
--- a/lib/carrierwave/locale/en.yml
+++ b/lib/carrierwave/locale/en.yml
@@ -9,6 +9,7 @@ en:
       content_type_allowlist_error: "You are not allowed to upload %{content_type} files, allowed types: %{allowed_types}"
       content_type_denylist_error: "You are not allowed to upload %{content_type} files"
       processing_error: "Failed to manipulate, maybe it is not an image?"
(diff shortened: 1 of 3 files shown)


mshibuya:
Makes sense, thank you!

PR #2824, opened 2026-07-31
Update image-processing to 2.x

woutersmeenk:
Similar to https://github.com/carrierwaveuploader/carrierwave/pull/2820, but with fix for CI.

PR #2820, opened 2026-06-09
Update image_processing dependency to 2.x

ciremoussadia:

PR #2826, opened 2026-08-05
Relax dependency on image_processing to allow updates to `~> 2.0`

### LINKED ISSUE #2825: Block unfuzzed loaders by default on libvips 8.13+
Hi!

I am looking at [[CVE-2026-66066] Possible arbitrary file read and remote code execution in Active Storage variant processing](https://discuss.rubyonrails.org/t/cve-2026-66066-possible-arbitrary-file-read-and-remote-code-execution-in-active-storage-variant-processing/91432) and seeing that `ruby-vips` is a transient dependency of `carrierwave`, so it might open the door to the same exploit.

There are several ways to go about disabling the unfuzzed loaders, so I am starting the discussion first and not opening any PRs yet:

1. Relax the dependency on `image_processing` to allow updates to version `>= 2.0.0` that includes https://github.com/janko/image_processing/commit/201586c8bbdfd9e17b445d3ad8601cb3fdd732a9.
2. Get upstream `image_processing` to release a maintenance `~> 1.1`  version with a cherry-pick of https://github.com/janko/image_processing/commit/201586c8bbdfd9e17b445d3ad8601cb3fdd732a9.
3. Add a redundant copy of the same code to [this block](https://github.com/carrierwaveuploader/carrierwave/blob/b5f0abe10ecf6500309fc83e0e8969cf57ba690e/lib/carrierwave/processing/vips.rb#L55-L59).
4. Any or all of the above.

WDYT?

amiryalsa:
Fixes #2825.

Related past work: #2824, #2820

### DIFF
--- a/carrierwave.gemspec
+++ b/carrierwave.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency "activesupport", ">= 7.0.0"
   s.add_dependency "activemodel", ">= 7.0.0"
-  s.add_dependency "image_processing", "~> 1.1"
+  s.add_dependency "image_processing", [">= 1.1", "< 3"]
   s.add_dependency "marcel", ">= 1.0.0", "< 2"
   s.add_dependency "addressable", "~> 2.6"
   s.add_dependency "ssrf_filter", "~> 1.0"
@@ -36,6 +36,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "fog-google", ["~> 1.7", "!= 1.12.1"]
   s.add_development_dependency "fog-local"
   s.add_development_dependency "mini_magick"
+  s.add_development_dependency "ruby-vips"
 
   if RUBY_ENGINE != 'jruby' && ENV['GITHUB_JOB'] != 'rubocop'
     s.add_development_dependency "rmagick", ">= 2.16"
--- a/spec/processing/mini_magick_spec.rb
+++ b/spec/processing/mini_magick_spec.rb
@@ -283,9 +283,7 @@ describe CarrierWave::MiniMagick do
     context "of being configured to use ImageMagick but failing to execute" do
       before do
         allow(MiniMagick).to receive(:processor).and_return(:magick)
-        allow(Open3).to receive(:capture3).and_raise(Errno::ENOENT)
         allow(Open3).to receive(:popen3).and_raise(Errno::ENOENT)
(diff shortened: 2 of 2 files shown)


julian-villarreal-cf:
The two `raise_exception` matchers in `spec/processing/mini_magick_spec.rb` lost their error class, and I think that ends up hiding the scenario those contexts were written to cover.

`MiniMagick::Shell#execute` rescues `Errno::ENOENT` internally and returns status 127; `#run` then raises `MiniMagick::Error` with an `executable not found: "…"` message. That's the path the old stub exercised — raising from `execute_open3` happened *inside* `execute`, so mini_magick wrapped it, and carrierwave's `raise e if e.message =~ /executable not found/` passthrough in `minimagick!` / `manipulate!` kicked in. Hence the original `MiniMagick::Error` / `minimagick_error` expectations.

Stubbing `#execute` itself raises *above* that rescue, so a bare `Errno::ENOENT` propagates and `rescue ::MiniMagick::Error, ::MiniMagick::Invalid` never fires. The examples still pass, but only because the matcher now accepts any exception — they no longer assert what their names say (`it "raises MiniMagick::Error"` / `it "raises MiniMagick::Invalid"`), and the `executable not found` passthrough is no longer covered. A bare `raise_exception` would also swallow e.g. a `NoMethodError` from a future refactor.

The underlying problem is real, though: `execute_open3` is gone in mini_magick 5 (the `popen3` call is inlined into `execute`), and 4.12 doesn't call `Open3.capture3` either — it uses `popen3` — so both of the old stubs were no-ops on 5.x. Stubbing one level lower works on both 4.x and 5.x:

```ruby
allow(Open3).to receive(:popen3).and_raise(Errno::ENOENT)
```

That keeps `require 'open3'` and lets `raise_exception(MiniMagick::Error)` and `raise_exception(minimagick_error)` stay exactly as they were, so the wrapping behaviour stays under test.

amiryalsa:
Thanks @julian-villarreal-cf! The fact that `capture3` calls `popen3` internally had slipped my attention, and I didn’t dig through to where `MiniMagic::Shell` had `rescue`d from `Errno::ENOENT`. Your exact suggestion (which I arrived at from first principles after reading your entire comment) seems to do the trick better.

Apologies to everyone also for [the misfired PR](https://github.com/carrierwaveuploader/carrierwave/pull/2827) earlier.

julian-villarreal-cf (approved):

mshibuya:
Thanks, but could you please revert the changes ‎in spec/processing/mini_magick_spec.rb‎?
It received a slightly different fix by #2828.

mshibuya:
Thank you!

PR #2807, opened 2026-02-21
upgrade marcel gem currently locked to '~> 1.0.0', lock to '~> 1.0'

braindeaf:
`marcel` is currently locked to `~>1.0.0` this PR updates the lock it to `~> 1.0` so that we can drag `marcel` up to the most recent version.

PR #2828, opened 2026-08-21
Relax marcel dependency

mkon:
Marcel is currently locked to 1.0.0 which causes some issue for us. I think most 1.x versions should be fine.

This can replace #2807 

I fixed one unrelated test that broke from `mini_magick` silently upgrading to `5.3.2`


mkon:
@mshibuya @plribeiro3000 can you take a look?

mshibuya:
Thank you!

mkon:
thank you as well for merging @mshibuya. Any chance for a patch release containing this change?

mshibuya:
Released as 3.1.4.

PR #2829, opened 2026-08-29
docs: add note on store_dir override for standalone uploader

akitoshiga:
Hi!
I realized the default `store_dir` depends on `model.id`. 
Adding a brief note about this requirement will help other developers avoid the same pitfall when using uploaders directly.

### Summary
Adds a note in the README explaining that standalone uploaders must override `store_dir`.

### Context & Motivation
When trying to use CarrierWave stand-alone (without mounting to a model), I ran into `NoMethodError: undefined method 'id' for nil` and got stuck troubleshooting for a while. 

Many Thanks!🙏

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,14 @@ class AvatarUploader < CarrierWave::Uploader::Base
 end
 ```
 
+Since the default `#store_dir` depends on a model, override it when using an uploader stand-alone to avoid NoMethodErroron model.id.
+
+```ruby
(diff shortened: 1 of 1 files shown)


mshibuya:
Thanks, but this is too specific to be included into the Getting Started section.
Please consider adding a Wiki how-to entry for standalone setup instead.
https://github.com/carrierwaveuploader/carrierwave/wiki#how-tos

